PR #2, opened 2017-02-14
ENH: simple graphviz transition digraphs

klauer:
Not too shabby

command_triggered_channel_transitions_server
command_triggered_channel_transitions_client
command_triggered_circuit_transitions_server
command_triggered_circuit_transitions_client


danielballan:
Ha! That's excellent. I'll merge and we'll iterate from there.

Taking (yet another) cue from h11, we might consider orphaning the Error state before passing the graph in. This would cause graphviz to set it off to the set, simplify the visuals. That fact that any state can lead to Error can be noted in a caption.

PR #6, opened 2017-03-05
ENH: add state_changed callback on channels

klauer:
This may not be how it was intended to be used - here for discussion.
Should instead the virtual circuit be monitoring what happens to the
channels instead of the channels themselves?

### DIFF
--- a/caproto/_hub.py
+++ b/caproto/_hub.py
@@ -233,8 +233,8 @@ class VirtualCircuit:
             # Update the state machine of the pertinent Channel.
             # If this is not a valid command, the state machine will raise
             # here.
-            chan._state.process_command(self.our_role, type(command))
-            chan._state.process_command(self.their_role, type(command))
+            chan._process_command(self.our_role, type(command))
+            chan._process_command(self.their_role, type(command))
 
             # If we got this far, the state machine has validated this Command.
             # Update other Channel and Circuit state.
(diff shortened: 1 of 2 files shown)


caproto/_state.py
```diff
@@ -188,10 +200,13 @@ def _fire_state_triggered_transitions(self, role):
...
             self.states[role], self.circuit_state.states[role] = new
+            return new
 
     def process_command(self, role, command_type):
-        self._fire_command_triggered_transitions(role, command_type)
-        self._fire_state_triggered_transitions(role)
+        command_trig = self._fire_command_triggered_transitions(role,
+                                                                command_type)
+        state_trig = self._fire_state_triggered_transitions(role)
+        return command_trig or state_trig
```
danielballan: What if a command triggers and transition (`command_trig` is not None) and then a state transition is triggered (`state_trig` is not None). Don't we want `state_trig` returned in that case?

caproto/_hub.py
```diff
@@ -586,6 +586,15 @@ def _fill_defaults(self, data_type, data_count):
             data_count = self.native_data_count
         return data_type, data_count
 
+    def state_changed(self, new_state):
```
danielballan: Is this just here in case a subclass wants to use it?

caproto/_state.py
klauer: You're right - swapped

caproto/_hub.py
klauer: Yes - modified docstring as note

danielballan: It might be useful to have the old state too. I usually see this done as `self, new_state, old_state`.

klauer:
One more attempt... I'm a bit rusty

danielballan:
Even Rusty Ken keeps me on my toes.

PR #11, opened 2017-03-18
DOC: Auto-build state machine graphs.

danielballan:
Auto-builds @klauer's beautiful state machine graphs and puts them into the docs.

### DIFF
--- a/doc/gen_graphs.py
+++ b/doc/gen_graphs.py
@@ -52,7 +52,7 @@ def generate(format_='png'):
             graph = create_transition_graph(d[role], role=role,
                                             format_=format_)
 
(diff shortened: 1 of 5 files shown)


danielballan:
FYI, @klauer, if you feel like falling down a deeper rabbit hole, take look at how Nathaniel incorporated the state-based transitions and added titles: https://github.com/njsmith/h11/blob/57d06346db936a2fc216464efe29c19cf7bbaf41/docs/source/make-state-diagrams.py

klauer:
Those state diagrams look really nice in h11. Will have to take a look into doing that for this project, too.

PR #16, opened 2017-03-21
ENH: mucking around with the server test

klauer:
Mucking around classifies as "ENH", ... right?

(note: this is on top of @danielballan's `test-server` branch)


danielballan:
Hahaha. I like the direction this is going in. When I have a dubious "ENH" I sometimes wish there were a standard abbreviation for "Pile of Technical Debt."

caproto/examples/curio_server.py
```diff
@@ -111,19 +233,17 @@ def __init__(self, host, port, pvdb):
                     res = ca.VersionResponse(13)
                     responses.append(res)
                 if isinstance(command, ca.SearchRequest):
-                    known_pv = command.name.decode() in self.pvdb
+                    known_pv = command.name.decode('latin-1') in self.pvdb
```
danielballan: I didn't come across this. Does EPICS use latin-1 as opposed to UTF-8?

klauer: It uses... well, bytes with no specific encoding.  UTF-8 certainly wasn't around that time, and I think other than ASCII (limited to 7-bits) latin-1 is probably closest to what was intended.

That said, we could potentially make this a configurable option. I wonder what the other implementations do?

klauer:
Closing in favor of #17

PR #17, opened 2017-04-09
PTD/TST: use CircleCI

klauer:
Sorry, TravisCI, you suck.

(*) PTD="Pile of Technical Debt"

TODO: handling of user-specified `EPICS_CA_ADDR_LIST` is not perfect

PR #18, opened 2017-04-10
CI: use CircleCI only, removing TravisCI

klauer:
Cherry-picked from #17, as a test to see if circle is building PRs

Should be updated to build docs before being used to fully replace travis.


danielballan:
I just realized this kills my docs-building integration with gh-pages. Not ideal, since the 'doctr' utility saves a bunch of hassle.

klauer:
Is there something that makes it not possible to integrate the docs-building integration with gh-pages on circle?

danielballan:
No, it's possible, but just more hassle since [doctr](https://github.com/drdoctr/doctr) targets Travis specifically. But I'd say it's worth it for the benefits of Circle-CI in this case.

PR #27, opened 2017-04-13
ENH: Remaining DBR types

klauer:
* DBR_GR_{*}
* DBR_STSACK_STRING
* DBR_CLASS_NAME

The framework was there, but not support in the server/payload-handling functions. This should cover the rest.

... questionable design decisions included as usual (and subject to negotiation)

... and failing travis due to `ca_test` not being there - will try again some other time.

### DIFF
--- a/caproto/_dbr.py
+++ b/caproto/_dbr.py
@@ -840,7 +840,7 @@ class ChannelType(IntEnum):
     TIME_LONG = 19
     TIME_DOUBLE = 20
 
-    GR_STRING = 21
+    GR_STRING = 21  # not implemented by EPICS
     GR_SHORT = 22
     GR_INT = GR_SHORT
     GR_FLOAT = 23
@@ -849,7 +849,7 @@ class ChannelType(IntEnum):
     GR_LONG = 26
     GR_DOUBLE = 27
 
-    CTRL_STRING = 28
+    CTRL_STRING = 28  # not implemented by EPICS
     CTRL_INT = 29
     CTRL_SHORT = 29
     CTRL_FLOAT = 30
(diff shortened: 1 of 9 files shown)


tacaswell:
I wonder if ccache + travis caching + just build epics-base would save any time?

Of course that means understanding epics build well enough to run it on travis.

klauer:
epics-base is rather easy to build. Once you get into the realm of trying to build the motor module and areaDetector and such on top of it is where you get into trouble.

tacaswell:
it would also let us get out of the sudo required pool which gets faster starts and (I think) better machines.

hhslepicka:
I have being doing this for quite a while here at work! Let me know if I
can be of any help! :)

tacaswell:
@hhslepicka can you make travis build epics-base?  We probably want to (eventually) do a matrix of libca versions.

hhslepicka:
@tacaswell I can do that! 👍

danielballan:
It looks like catvs runs Travis against a matrix of different branches of epics-base. See https://github.com/danielballan/catvs/blob/master/build-deps.sh and the .travis.yml.

Also, that repo has no setup.py... maybe we can send a PR in return for the free help.

klauer:
Rebased this old one, but it will still need some work.  I don't think it'd be wise for me to do it with how my head feels, so I'll leave this for another day.

danielballan:
Force-pushed another rebase.

caproto/_dbr.py
```diff
@@ -1072,7 +1098,9 @@ def promote_type(ftype, *, use_status=False, use_time=False, use_ctrl=False):
         ftype += ChType.TIME_STRING
     elif use_status:
         ftype += ChType.STS_STRING
+    elif use_gr:
```
tacaswell: This code block should get some comment about the assumption that the types are in blocks, all in the same order, and than the string type is the first one in each block.

danielballan: Good point -- it took me a second to grok that too. Addressed.

danielballan:
Well, we now support all the DBR types. :tada:

PR #28, opened 2017-04-14
ENH: CA status codes

klauer:


caproto/_utils.py
```diff
@@ -69,10 +70,6 @@ class RemoteProtocolError(ProtocolError):
     ...
 
 
-class CaprotoKeyError(CaprotoError, KeyError):
```
tacaswell: What you can not make up your mind on MRO order!

tacaswell:
What do you do with these things now that we have them defined?

klauer:
From the docs: 
> Return codes are communicated in the protocol by the CA_PROTO_READ_NOTIFY, CA_PROTO_WRITE_NOTIFY, monitor subscription responses, and the CA_PROTO_ERROR responses.

For such things as 'access denied' and 'bad channel ID' - we should be responding with appropriate errors instead of bailing out, eventually.

caproto/_utils.py
klauer: _Seriously_, @danielballan!

klauer:
Thanks, @danielballan. I preferred how this [other branch](https://github.com/klauer/caproto/commit/ff0df2f7964fa4dc4a8fa0142db98bbadef74779) looked, but couldn't quite get it into a nice final solution.

PR #29, opened 2017-04-15
TST: start of basic benchmarking comparison

klauer:
Only a start, more useful for the test utilities (on-the-fly softIoc execution) at this point


danielballan:
Have you ever played around with asv? I think that what we need now (a quick comparison) doesn't necesaarilt call for something as heavy as asv but this PR is well on its way to reinventing asv.

klauer:
I haven't yet, but if you think it would be worthwhile we could certainly give it a go!

danielballan:
Superceded by #35 where all this nice work is refactored into asv benchmarks. Thanks, @klauer!

PR #31, opened 2017-04-18
CI/DOC: Run repeater for docs build and fail on docs warnings.

danielballan:


danielballan:
Self-merging this docs-only change to fix broken docs.

PR #32, opened 2017-04-18
Remove extraneous transitions and add missing ones.

danielballan:
The state machines needed some code cleanup for readability and
allowed some transitions that in fact were not legal (artifacts of
early misunderstandings in the design process).

Amazingly, all the tests still pass in one try!
I sure hope Travis agrees.

### DIFF
--- a/caproto/_state.py
+++ b/caproto/_state.py
@@ -178,6 +178,8 @@ STATE_TRIGGERED_TRANSITIONS = {
 
 
 class _BaseState:
+    TRANSITIONS = {}  # Subclasses should override this with non-empty dict.
+
     def __repr__(self):
         return "<{!s} states={!r}>".format(type(self).__name__, self.states)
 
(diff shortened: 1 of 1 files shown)


caproto/_state.py
```diff
@@ -22,70 +22,77 @@
 COMMAND_TRIGGERED_CIRCUIT_TRANSITIONS = {
     CLIENT: {
         SEND_VERSION_REQUEST: {
-            EchoRequest: SEND_VERSION_REQUEST,
-            EchoResponse: SEND_VERSION_REQUEST,
             VersionRequest: AWAIT_VERSION_RESPONSE,
-            VersionResponse: CONNECTED,
```
danielballan: A lot of complexity in the states came from bouncing directly to CONNECTED, which is in correct in my view. If the CLIENT is in a SEND_VERSION_REQUEST state, it should still send a version request (and then get a response in turn).

caproto/_state.py
```diff
@@ -178,23 +202,29 @@
...
+
     def __repr__(self):
         return "<{!s} states={!r}>".format(type(self).__name__, self.states)
 
     def __getitem__(self, role):
         return self.states[role]
 
     def _fire_command_triggered_transitions(self, role, command_type):
-        state = self.states[role]
+        current_state = self.states[role]
```
danielballan: Separating this into two lines isolates the KeyError.

danielballan:
N.B. Codecov is doing something weird but the /patch coverage is happy and so is Travis.

klauer:
I gave this a read through finally and I believe it to be correct, and as such I will take 100% of 0 responsibility when we poke holes in it later.

danielballan:
<3

PR #35, opened 2017-04-19
Set up asv for benchmarking.

danielballan:
Using [asv](https://asv.readthedocs.io/en/latest/) 

Early results, which might be based on errors but are at least promising:

```
[ 25.00%] ··· Running benchmarks.TimeWaveform4000.time_caproto_curio            105.61ms
[ 50.00%] ··· Running benchmarks.TimeWaveform4000.time_pyepics                   56.68ms
[ 75.00%] ··· Running benchmarks.TimeWaveformMillion.time_caproto_curio         110.12ms
[100.00%] ··· Running benchmarks.TimeWaveformMillion.time_pyepics               342.11ms
```


klauer:
Some tweaks in [https://github.com/klauer/caproto/commits/simple_bench] which I encountered while trying to give asv a go. Overall I like it, but I'm not sure how best to use it.

How difficult will it be to get this hosted along with the docs?

danielballan:
I merged your commits into this branch. It's quite easy to push this up alongside the docs: they are now published at https://danielballan.github.io/caproto/bench

If I understand the operation of doctr correctly, subsequent docs updates should not blow those away. Eventually we should pick a standard machine to execute these on -- likely frejya -- and publish manually or through a cronjob.

danielballan:
These @klauer's commits + some new asv-related files from me that don't touch any package code. Merging so we can build benchmarks on this in parallel PRs.

PR #40, opened 2017-04-20
TST/FIX: Serialization round trip

klauer:
A simple test with a handful of fixes found already.

Payload comparison is failing currently (and is an important TODO for this PR).

### DIFF
--- a/tests/test_serialization.py
+++ b/tests/test_serialization.py
@@ -3,75 +3,82 @@ import socket
 
 import caproto as ca
 import inspect
+import pytest
 
+# TODO this is used in several places in _commands.py
+ip = '255.255.255.255'
+encoded_ip = socket.inet_pton(socket.AF_INET, ip)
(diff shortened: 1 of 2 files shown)


tests/test_serialization.py
```diff
@@ -0,0 +1,77 @@
...
+
+        ba = sig.bind(**bind_args)
+
+        print(cmd, ba.arguments)
+        inst = cmd(*ba.args, **ba.kwargs)
+
+        if cmd in (ca.Message, ):
+            continue
+
+        role = (ca.CLIENT if cmd.__name__.endswith('Request')
```
danielballan: Use `ca.DIRECTION` (which is `REQUEST` or `RESPONSE`)

danielballan:
Yay it passes! It may be worth giving some more thought to optimizing this __bytes__ method because it is a hot path, but I think we can defer than until we actually do some measurements and see where the slow lines are.

If you think this approach is reasonable @klauer I leave it to you to merge.

PR #34, opened 2017-04-19
ENH: working threaded implementation

tacaswell:
can register and search (so far)

PR #44, opened 2017-04-21
Test PV drop-in class

tacaswell:
This builds on #34 and re-uses tests from pyepics


tacaswell:
df82ee37aa33d8032ce1d6013bdeb2325ed47ff2 is the source of the bug.  Reverting that fixes the serialization tests and breaks the pv tests.

danielballan:
Merging mainly to get the hub changes in. Tom wants to continue this in a separate PR anyway. Everyone wins!

PR #45, opened 2017-04-21
REF: move up user-friendlyish data-handling classes

klauer:
TODOs
- [x] All tests passing
- [ ] Make more tests
- [x] Does extract_data slice copy? No: [All arrays generated by basic slicing are always views of the original array.](https://docs.scipy.org/doc/numpy/reference/arrays.indexing.html)
- [x] Fix docs generation failure in CI so PR author does not go mad [_too laaaaate_ - @danielballan]
- [ ] Fix crappy @klauer code
- [ ] Following the previous point, refactor "DatabaseRecord" naming and implementation into something cleaner/neater/more useful/performant
- [ ] User-facing attributes should not be bytes but str (`units` in `_data.py`)
- [ ] Address `data_count` discrepancy (see asv benchmark)
- [ ] @tacaswell discovered issues with the threading client?


caproto/_commands.py
```diff
@@ -99,60 +99,28 @@ def padded_string_payload(payload):
...
-        value_payload = b''.join(map(bytes, values))
+def data_payload(data, metadata, data_type, data_count):
+    if isinstance(data, ctypes.BigEndianStructure):
+        # This is already a DBR.
+        encoded_data = data
+    else:
+        # This is a tuple of values to be encoded into a DBR.
+        ntype = native_type(data_type)
+        cls_dict = {'_fields_': [('value', ntype * data_count)]}
+        cls = type(name, (ctypes.BigEndianStructure,), cls_dict)
```
klauer: Structures aren't meant for dealing with homogeneously-typed arrays alone.
It's more the job of `struct.pack`:
```python
import struct
struct.pack('>{}i'.format(data_count), *int_array)
```
Or the Python standard `array`, which I've never used somehow: [python docs](https://docs.python.org/3/library/array.html#module-array)
```python
from array import array
arr = array('l', data)
if little_endian:
    arr.byteswap()
```

caproto/_commands.py
```diff
@@ -99,60 +99,28 @@ def padded_string_payload(payload):
...
+        cls(data)  # This is probably wrong in a super dumb way.
+    if isinstance(metamdata, ctypes.BigEndianStructure):
```
klauer: So `values` here is the array, and `metadata` is the corresponding `DBR_TYPE` to the request, or a namedtuple/etc

caproto/_commands.py
```diff
@@ -99,60 +99,28 @@ def padded_string_payload(payload):
...
-                    #                  ''.format(attr, value, ex))
-                    print('set metadata fail', attr, value)
-    payload.value = value_payload.value
-
-    payload = bytes(payload)
+        # This is a tuple of values to be encoded into a DBR.
+        justified_md = []
+        for val in metadata:
+            if isinstance(val, str):
+                val = bytes_(val).ljust(MAX_ENUM_STRING_SIZE, b'\x00')
```
klauer: `len(val)` could be > `MAX_ENUM_STRING_SIZE`, then if so I guess the DBR_TYPES[]() initializer bails or chops off the rest?

caproto/_commands.py
```diff
@@ -1122,37 +1103,34 @@ class ReadNotifyResponse(Message):
...
+    def __init__(self, data, metadata, data_type, data_count, status, ioid):
+        size, payload = data_payload(data, metadata, data_type, data_count)
         header = ReadNotifyResponseHeader(size, dbr_data.DBR_ID, data_count,
                                           status, ioid)
         super().__init__(header, payload)
 
     payload_size = property(lambda self: self.header.payload_size)
+    data = property(lambda self: extract_data(self.payload, self.data_type,
+                                              self.data_count))
+    metadata = property(lambda self: extract_metadata(self.payload,
```
klauer: What if we'd rather work with the `DBR_` type for metadata?

caproto/_dbr.py
```diff
@@ -120,7 +120,6 @@ class DBR_STS_STRING(ctypes.BigEndianStructure):
     _fields_ = [
         ('status', short_t),
         ('severity', short_t),
-        ('value',  string_t),
```
klauer: These will lead to more consistent scalar/array handling. 👍

caproto/_commands.py
klauer: `memoryview` and `memoryview.cast` may also be of interest

danielballan: I've been on the lookout for an excuse to use `array`. These are equivalent style-wise, so we should let performance break the tie. There may be little difference.

danielballan:
> i'd like to get this whole PR passing tests, then merge, then refactor from there?

- @klauer on GChat

In that spirit, I'm declaring victory, but let's keep track of the TODOs at the top of this PR.

PR #46, opened 2017-04-21
FIX/TST: Various fixes and tests to increase coverage of _hub.py

danielballan:


danielballan:
Self-merging to get bug fixes in master. Nothing interesting here.

PR #50, opened 2017-04-26
TST: More test coverage with sundry fixes and a couple API changes

danielballan:


caproto/_hub.py
```diff
@@ -885,14 +885,14 @@ def clear(self):
         command = ClearChannelResponse(self.sid, self.cid)
         return command
 
-    def read(self, values, ioid, data_type=None, data_count=None, status=1, *,
+    def read(self, data, ioid, data_type=None, data_count=None, status=1, *,
```
tacaswell: *grumble* api change *grumble

danielballan: This kind of grumbling is frowned on in caproto.

PR #51, opened 2017-04-26
FIX: Parse CHAR waveform as bytes not numpy array of chars.

danielballan:
I think this is what we agreed on yesterday (and it's what is explained in the comments for this function) but somehow the implementation got lost. All the same tests pass; this is just a quality-of-life measure.


danielballan:
@tacaswell This probably broke some of your spaghetti-spaghetti.

danielballan:
Locally the two failures I can reproduce are:

```
tests/test_thread_pv.py::PV_Tests::test_emptyish_char_waveform_monitor
tests/test_thread_pv.py::PV_Tests::test_emptyish_char_waveform_no_monitor
```

PR #52, opened 2017-04-26
CI: allow py.test to capture stdout/stderr due to current verbosity

klauer:
Dying-on-the-couch PR numero 1


hhslepicka:
👍 for the "numero"

danielballan:
Caproto enthusiastically supports multiculturalism in the development process.

PR #57, opened 2017-04-26
API/TST: Test channel creation failure, disconnect, and clearing.

danielballan:
This includes some fixes and changes to the state machine:

- There is no need for an ERROR state. Remove it.
- We do need a state to go when we see CreateChFailResponse. Call it FAILED.
- Handle CreateChFailResponse properly in the VirtualCircuit.
- Add a `create_fail()` convenience method on `ServerChannel` that composes a `CreateChFailResponse` and a `disconnect` method that composes a `ServerDisconnResponse`.
- Add a `disconnect()` method on `VirtualCircuit` that updates its state machine.


danielballan:
There is more cleanup and testing I can add to this PR before it is merged.

caproto/_hub.py
```diff
@@ -299,6 +299,15 @@ def _process_command(self, role, command):
                           "of {} for this circuit".format(command.priority,
                                                           self.priority))
 
+    def disconnect(self):
+        """
+        Notify all channels on this circuit that they are disconnected.
+
+        Clients are servers should call this method when a TCP connection is
```
tacaswell: clients _and_ servers?

caproto/curio/client.py
```diff
@@ -44,7 +44,7 @@ def __init__(self, circuit):
             raise RuntimeError("must await create_connection() first")
         bytes_received = await self.socket.recv(32768)
         if not len(bytes_received):
-            raise ca.DisconnectedCircuit()
```
tacaswell: should this return here?  not sure throwing 0 bytes into circuit is correct now that the circuit thinks it is disconnected.

danielballan: good catch

caproto/_hub.py
danielballan: Just clients. _Reaches for more coffee..._

caproto/_state.py
```diff
@@ -171,32 +169,42 @@
...
+            ErrorResponse: FAILED
         },
     },
 }
 
 STATE_TRIGGERED_TRANSITIONS = {
     # (CHANNEL_STATE, CIRCUIT_STATE)
     CLIENT: {
-        (NEED_CIRCUIT, CONNECTED): (SEND_CREATE_CHAN_REQUEST, CONNECTED),
+        (SEND_CREATE_CHAN_REQUEST, DISCONNECTED): (CLOSED, DISCONNECTED),
```
tacaswell: These seem to only have one end point, can we squash some complexity?

danielballan: Potentially. I want to leave this here for now because it might be useful later and I'm not sure we're done fiddling with it.

caproto/_state.py
```diff
@@ -257,6 +265,12 @@ def process_command_type(self, role, command_type):
         for chan in self.channels.values():
             chan.states._fire_state_triggered_transitions(role)
 
+    def disconnect(self):
+        self.states = {CLIENT: DISCONNECTED, SERVER: DISCONNECTED}
+        # Notify channels on this circuit.
+        for chan in self.channels.values():
+            chan.states._fire_state_triggered_transitions(CLIENT)
```
tacaswell: does the order that this happens in matter?

Touching private state on attributes is a bit of tight coupling (what are they, c++ friends?), would it be better to have a single `update_state` public method on the channels?

danielballan: Easy enough. Done.

PR #59, opened 2017-04-27
FIX: disconnect if socket closes

tacaswell:


caproto/threading/client.py
```diff
@@ -391,8 +427,28 @@ def __init__(self, pvname, callback=None, form='time',
...
         # not handling lazy instantiation
         self._context.search(self.pvname)
+        self.chid = None
+        self.wait_for_connection(form=form, count=count)
+
+    @property
+    def connected(self):
+        return self.chid is not None and self.chid.connected
+
+    def force_connect(self, pvname=None, chid=None, conn=True, **kws):
```
danielballan: I'm interested where you're going with this....

tacaswell: that is a hold over from pyepics.

caproto/_state.py
```diff
@@ -180,6 +181,7 @@
             ErrorResponse: MUST_CLOSE,
         },
         CLOSED: {
+            ClearChannelResponse: CLOSED,
```
danielballan: This seems unnecessarily tolerant to me, especially because it's a *response*.

tacaswell: thrashing to try to fix it ;)

tacaswell: This is probably not the right fix anyway,  there seems to be a race condition when things are torn down where the channel is closed twice.

tacaswell: and this copies the behavior in `FAILED` below.

danielballan:
I forgot to submit that review from the train. I pushed a commit to address my own comments.

PR #63, opened 2017-04-28
TST: Test that all the states are what we expect through a transaction.

danielballan:
No surprises here. Extra tests for due diligence.


danielballan:
Just adding tests. Merging summarily.

PR #69, opened 2017-05-01
Fix repeater

tacaswell:
And some other changes.


danielballan:
Thanks for digging into this, @tacaswell.

PR #71, opened 2017-05-02
DOC: add some TODOs in the code

tacaswell:


caproto/threading/client.py
```diff
@@ -546,15 +550,20 @@ def connected(self):
         return self.chid is not None and self.chid.connected
 
     def force_connect(self, pvname=None, chid=None, conn=True, **kws):
+        # not quite sure what this is for in pyepics, probably should
+        # be an arias for reconnect?
```
danielballan: Maybe it's my latent musical education, but I'm not sure "arias" means other than singing while waiting for the PV to connect.

tacaswell: a miss type of 'alias' (r and l are next to each other in dvorak on o and p).

However if you were required to sing while waiting for PVs to connect, it might provide motivation to increase reliability.

PR #72, opened 2017-05-02
ENH/TST: Support put completion in threading client.

danielballan:
This fixes an accidental API mismatch with pyepics: get() and put() should return ``command.data``, not ``command``.

### DIFF
--- a/tests/test_examples.py
+++ b/tests/test_examples.py
@@ -140,8 +140,8 @@ def test_thread_pv():
     print('time read', time_pv.get())
     print('ctrl read', ctrl_pv.get())
 
-    time_pv.put(3)
-    time_pv.put(6)
+    time_pv.put(3, wait=True)
+    time_pv.put(6, wait=True)
 
     time.sleep(0.1)
     assert time_pv.get() == 6
--- a/caproto/threading/client.py
+++ b/caproto/threading/client.py
@@ -295,9 +295,10 @@ class VirtualCircuit:
 
                 if isinstance(command, ca.ReadNotifyResponse):
                     chan = self.ioids.pop(command.ioid)
-                    chan.last_reading = command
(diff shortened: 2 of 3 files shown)


tests/test_examples.py
```diff
@@ -140,8 +140,8 @@ def user_callback(*, value, **kwargs):
     print('time read', time_pv.get())
     print('ctrl read', ctrl_pv.get())
 
-    time_pv.put(3)
-    time_pv.put(6)
+    time_pv.put(3, wait=True)
```
danielballan: This was working previously because `wait=False` (the default) actually waited.

tacaswell: You are too kind.  'This worked previously because some yahoo ignored the wait kwarg' :stuck_out_tongue_winking_eye:

tacaswell:
Would it be better to push the wait up into `Channel.write` and not have to use an extra thread?

danielballan:
I don't see any threading code in `Channel.write`. Do you mean push it all the way up through `Channel.write` and further into `circuit.send`?

tacaswell:
Pick the wait kwarg out and do something like:

```python
        if not wait:
            return None
        while True:
            with self.circuit.has_new_command:
                if ioid not in self.circuit.ioids:
                    break
                if not self.circuit.has_new_command.wait(60):
                    raise TimeoutError()
        return self.last_reading

```

or make a `awrite` that returns a future and have `write` call that and from `write`.

danielballan:
I now see while `_read_notify_callback` and `_write_notify_callback` still exist, they are no longer involved in handling put-completion callbacks and there is no race condition anymore. I'm not sure why we need them -- they don't seem to be publicly accessible, directly or indirectly -- but they aren't doing any harm and could be useful later.

PR #73, opened 2017-05-06
REF: use queues at a lower level

klauer:
- [x] curio.client
 - [x] curio.server
 - [x] threading.client
 - [x] most tests
 - [x] synchronous client (less than ideal)
 - [x] repeater (less than ideal)
 - [x] docs building
 - [x] locking and timeout re-work for threading client

There some additional junk as it's impossible for me to make a single-issue PR:

 - Concept of a shared broadcaster in the thread client (and at least moved out of context in curio client now)
 - Split out constants and broadcaster to separate file from hub

Further TODOs:
   - Further split up `recv()` and command parsing (maximize throughput of socket)
   - ab43f05 was the start of attempting to do less Header/Message creation when not enough bytes are available
    - Special handling of next_command replacement slightly differently in different tests is not good
   - Blame @klauer for breaking everything

### DIFF
--- a/caproto/threading/client.py
+++ b/caproto/threading/client.py
@@ -121,7 +121,7 @@ class SharedBroadcaster:
             for host in ca.get_address_list():
                 self.udp_sock.sendto(bytes_to_send, (host, port))
 
-    def register(self):
+    def register(self, *, wait=True):
         "Register this client with the CA Repeater."
         if self.registered:
             return
@@ -133,7 +133,7 @@ class SharedBroadcaster:
             command = self.broadcaster.register('127.0.0.1')
             self.send(ca.EPICS_CA2_PORT, command)
 
-        while True:
+        while wait:
             with self.command_cond:
                 if self.registered:
                     break
@@ -495,7 +495,7 @@ class Channel:
                 if not self.circuit.new_command_cond.wait(2):
                     raise TimeoutError()
 
-    def disconnect(self):
+    def disconnect(self, *, wait=True):
         "Disconnect this Channel."
         with self.circuit.new_command_cond:
             if self.connected:
@@ -506,7 +506,7 @@ class Channel:
                     return
             else:
                 return
-        while True:
+        while wait:
             with self.circuit.new_command_cond:
                 if self.channel.states[ca.CLIENT] is ca.CLOSED:
                     break
--- a/caproto/threading/client.py
+++ b/caproto/threading/client.py
@@ -287,6 +287,10 @@ class Context:
                 circuit.send(cachan.version(),
                              cachan.host_name(),
                              cachan.client_name())
+                # NOTE: starting the response command evaluation thread here,
+                # after having sent the VersionRequest, fixes a race condition
+                # with handling of rsrv sending VersionResponse upon connection
+                circuit.command_thread.start()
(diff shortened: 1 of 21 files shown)


caproto/threading/client.py
```diff
@@ -41,71 +47,106 @@ def __call__(self):
             if target is None:
                 break
 
-            if len(bytes_recv):
-                target.next_command(bytes_recv, address)
-                # this is important to not keep a hard-ref to the target
-                target = None
-            else:
+            # this is important to not keep a hard-ref to the target
```
tacaswell: this comment needs to be on the `target = None` line.

caproto/threading/client.py
```diff
@@ -120,14 +161,73 @@ def search(self, name):
             self.send(ca.EPICS_CA1_PORT, ver_command, search_command)
         # Wait for the SearchResponse.
         while True:
-            with self.cntx_condition:
-                if search_command.cid not in self.unanswered_searches:
-                    break
-                if not self.cntx_condition.wait(2):
-                    raise TimeoutError('failed to find PV {}'.format(name))
+            with self.command_cond:
```
tacaswell: should this be the search lock?

tests/test_serialization.py
```diff
@@ -266,3 +267,46 @@ def test_bytelen():
...
     with pytest.raises(ca.CaprotoValueError):
         ca.SearchRequest(name='a' * 41, cid=0, version=13)
+
+
+skip_ext_headers = [
+    'MessageHeader',
+    'ExtendedMessageHeader',
+
+    # TODO: these have no args, and cannot get an extended header
+    'EchoRequestHeader',
```
klauer: Need to address these @danielballan

danielballan:
Wow, @klauer. This is a big step forward. I'd like to think about `next_command` and `process_command` a little more before we merge.

danielballan:
Caproto master:

* `recv` -> cache bytes but do no interpretation whatsoever
* `next_command` -> parse out one command's worth of bytes and update state

This PR:
* `recv` -> cache bytes and partition them into commands (and put those into a queue)
* `process_command` -> update state

What I would propose:
* `recv` -> same as this PR, maybe
* `next_command` -> Get one command from the queue and update state. Because this wraps `self.command_queue.get()` it will need sync and async variants.
* `backlog` -> a property showing the number of commands stashed

tacaswell:
How did _adding tests_ decrease the patch coverage?!

klauer:
Looking back on the proposed:
> next_command -> Get one command from the queue and update state. Because this wraps self.command_queue.get() it will need sync and async variants.

Does this mean that curio has to become a core dependency? (as the relevant code is in `caproto._hub.VirtualCircuit`) Or were you thinking that we have 2 separately-named methods (i.e., not using the curio magic): one that does `await queue.get()` and one that does `queue.get()`, requiring the caller to know which should be used?

I'm willing to make the changes, but I'd like to get them right this time as there's quite a bit of code dependent on these core methods.

klauer:
All but 1 test passing on 3.5 (`tests/test_threading_client.py::test_context_disconnect`), and I can't reproduce it locally...

klauer:
Superseded by #74 as everyone seems to agree with this PR's sentiment

PR #81, opened 2017-06-25
CI: try CircleCI 2.0 with custom primary docker container

klauer:
Reference:
* https://github.com/klauer/caproto-docker
* https://hub.docker.com/r/klauer/caproto-docker/

Docker image is still messy because it was made hastily, but it works!

Remaining things that need to be done that I won't be implementing in the near future: 
* matrix of Python versions (eventually epics-base versions as well?)
* doc building/pushing


danielballan:
Can you switch on Circle-CI in the repo settings? I have tried to do this on two projects and somehow put the project into a weird, broken state.

PR #83, opened 2017-06-25
WIP: use select() in threaded client

klauer:
Reduces number of threads in threaded client drastically. Can be reduced further in the future.

 - [ ] threadpool to handle message parsing
 - [x] fix up weakref failed attempt
 - [x] make the selector per-context and not per-process


klauer:
Tests passing again, finally

danielballan:
This should be rebased soon so that said rebase is not horrible.

klauer:
I'm not so thrilled about continuing work on this PR, if I'm being honest. I think it's probably still the right approach (or closer to it at least - the true right approach would be mirroring how EPICS base handles its potentially hundreds of connections)

danielballan:
Completely fair. I could be up for pushing this across the finish line. I'd like to understand `select` better.

danielballan:
Superceded by #124

PR #84, opened 2017-06-25
CLN: Misc cleanup on middle layers in prep for better docs

danielballan:
- Consistently follow pattern of naming things CurioVirtualCircuit etc.
- Move all testing code to tests, not in __main__ of modules. (Some tests were
  being maintained redundantly in both places; some needed to be moved.)

### DIFF
--- a/caproto/curio/client.py
+++ b/caproto/curio/client.py
@@ -3,8 +3,8 @@
 # It builds on the abstractions used in caproto, adding transport and some
 # caches for matching requests with responses.
 #
-# VirtualCircuit: has a caproto.VirtualCircuit, a socket, and some caches.
-# Channel: has a VirtualCircuit and a caproto.ClientChannel.
+# CurioVirtualCircuit: has a caproto.VirtualCircuit, a socket, and some caches.
+# CurioChannel: has a CurioVirtualCircuit and a caproto.ClientChannel.
 # Context: has a caproto.Broadcaster, a UDP socket, a cache of
 #          search results and a cache of VirtualCircuits.
 #
@@ -21,14 +21,14 @@ class ChannelReadError(Exception):
     ...
 
 
-class VirtualCircuit:
+class CurioVirtualCircuit:
     "Wraps a caproto.VirtualCircuit and adds transport."
     def __init__(self, circuit):
         self.circuit = circuit  # a caproto.VirtualCircuit
-        self.channels = {}  # map cid to Channel
-        self.ioids = {}  # map ioid to Channel
+        self.channels = {}  # map cid to CurioChannel
+        self.ioids = {}  # map ioid to CurioChannel
         self.ioid_data = {}  # map ioid to server response
-        self.subscriptionids = {}  # map subscriptionid to Channel
+        self.subscriptionids = {}  # map subscriptionid to CurioChannel
         self.connected = True
         self.socket = None
         self.command_queue = curio.Queue()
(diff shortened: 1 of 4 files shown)


danielballan:
This is not ready for review.

tacaswell:
To enter the :bike: :house_with_garden:  on this, would it be a better convention to in the mid layers `import caparto as ca` and always use the lower level classes as `ca.FOO` instead of having `caprto.curio.CurioFoo` (see https://youtu.be/o9pEzgHorH0?t=10m29s).

klauer:
Makes sense, though I admit I was partial to being able to run those examples individually and readily through `__main__` without py.test getting in the way.

tacaswell:
we could move the tests into the modules?  I also love `-k thread` which only runs the tests with the word 'thread' in their name.

klauer:
Yeah, I use `-k` often. I guess mostly I liked the curio server one, where I could easily poke around with `caget`, `caput`, and `camonitor` with a set of known PVs. 

Can we include some separate examples, now that the threaded/curio/async implementations are included in caproto base? I think the original argument against those was when they weren't intended to be part of the core library.

danielballan:
> Can we include some separate examples [...]?

Yeah, I had the same thought. I think that will address your needs and is a nice thing to provide anyway.

klauer:
Rebased and force-with-lease-pushed this old PR.  Examples were moved out in another PR, so this is only with renames.

Addressing @tacaswell's old bikesheddy comment, I dislike relying on the full module.submodule.classname as we have _many_ duplicate class names in caproto - oftentimes where one is using the other. I think it's justified to name them differently in this case.

danielballan:
Let’s not merge this just yet. Still want to mull over our options.

tacaswell:
I think this comes back to the same issue as our plans discussion, if you always do `import curio.Submodule as cS` instead of `from curio.Submodule import thing1, thing2` the ambiguity is eliminated.

I think there is value (other than my sense of aesthetics) to having the same class names in all of the middle layers.  It will simplify doing something like  https://github.com/NSLS-II/ophyd/blob/91ddd92be35b84607345f8c8af9e561a0bc2fb78/ophyd/signal.py#L41-L42 in a more generalized way (the auto quote does not seem to work between repos :( , it is the `cl` injection into `Signal`) to also expect more async co-routines (not sure exactly what those will look like yet).  We could also have each middle layer expose a `SimpleNamespace` with consistent names.

With `curio` and `threading` there are not many cases where we want to swap them out for each other, but I can imagine that with we will be able to write one `AsyncPV` class and swap out the asyncio/curio/trio backends for it.

danielballan:
I had also made that connection and was mentally composing a similar comment. The async and non-async variants are not drop-in replacements for one another, and while that makes this less compelling _now_, I think it will be more compelling later when we have multiple async variants in the codebase.

Here's an idea I'm not attached to (and may not even be in favor of)... Maybe instead of giving each transport implementation a special name, we should give the _core_ classes special names. I think the confusing part of this design comes from things like `context.circuit.circuit` and the fact that both `context.circuit` and `context.circuit.circuit` are instances of a class named `VirtualCircuit`. While a given application will rarely mix `curio.server.VirtualCircuit` with `asyncio.server.VirtualCirciuit`, _all_ applications will juggle some `<TRANSPORT>.<ROLE>.VirtualCircuit` and `caproto._core.VirtualCircuit`. Both will be commonly handled in interactive debugging sessions. Giving the latter a special name might help.

klauer:
Good discussion above, but this is an old and defunct PR. Closing.

PR #85, opened 2017-06-27
BENCH/WIP: benchmarking update

klauer:
* Separate IOC that needs to be run alongside asv (`python -m benchmarks.bench_ioc`)
* Current tests:
    - 4k array
    - 1M array
    - 2M array

Not satisfied with this layout just yet, need to rethink...

### DIFF
--- a/benchmarks/bench_ioc.py
+++ b/benchmarks/bench_ioc.py
@@ -1,0 +1,59 @@
+import os
+from . import util
+
+
+class IocHandler:
+    def __init__(self):
+        # NOTE: have to increase EPICS_CA_MAX_ARRAY_BYTES if NELM >= 4096
+        #       (remember default is 16384 bytes / sizeof(int32) = 4096)
+        self._cms = []
+        self._softioc_processes = []
+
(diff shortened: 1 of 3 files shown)


benchmarks/bench_ioc.py
```diff
@@ -0,0 +1,59 @@
...
+        self._cms = []
+        self._softioc_processes = []
+
+    def setup_ioc(self, *, db_text, max_array_bytes=16384, env_vars=None,
+                  **kwargs):
+        if env_vars is None:
+            env_vars = {}
+
+        env = dict(EPICS_CA_MAX_ARRAY_BYTES=str(max_array_bytes))
+        os.environ['EPICS_CA_MAX_ARRAY_BYTES'] = str(max_array_bytes)
```
danielballan: I know this is WIP -- just flagging this for later: lines 17 and 18 look redundant.

klauer: `os.environ` doesn't get passed to the subprocess, so that's why `env` is here. But good point, this line isn't really necessary.

klauer:
Closed in favor of #112 or anything else, really

PR #90, opened 2017-07-01
Refactoring _data.py

danielballan:
Just getting started here -- do not merge.

Next, I plan to fix up `ChannelData.set_dbr_data` (currently it seems to ignore its `metadata` parameter). Then I plan to refactor `ChannelData.__init__` to use `set_dbr_data`.

### DIFF
--- a/caproto/_data.py
+++ b/caproto/_data.py
@@ -210,6 +210,7 @@ class ChannelData:
         self.string_encoding = string_encoding
         self.reported_record_type = reported_record_type
         self._subscription_queue = None
+        # a dict of un-filled structs for possible future use
         self._dbr_metadata = {
             chtype: DBR_TYPES[chtype]()
             for chtype in (promote_type(self.data_type, use_ctrl=True),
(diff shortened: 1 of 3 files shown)


caproto/_data.py
```diff
@@ -236,18 +231,18 @@ def get_dbr_data(self, type_):
...
+        except KeyError:
             dbr_metadata = DBR_TYPES[type_]()
+            self._dbr_metadata[type_] = dbr_metadata  # cache for reuse
 
         self._copy_metadata_to_dbr(dbr_metadata)
         return dbr_metadata, values
 
     async def set_dbr_data(self, data, data_type, metadata):
         '''Set data from DBR metadata/values'''
+        self.timestamp = time.time()
```
klauer: Timestamp should be set after updating the value, as `fromtype` may reject it

caproto/_data.py
```diff
@@ -146,7 +146,8 @@ def string_encoding(self):
 
 
 class ChannelData:
-    data_type = ChType.LONG
+    # subclass must define a `data_type` class attribute, set to a value of the
+    # ChType enum
 
     def __init__(self, *, value=None, timestamp=None, status=0, severity=0,
                  string_encoding='latin-1', alarm_status=None,
```
danielballan: Notes to self:
* `values` should be called `data` for consistency with everything else
* `string_encoding` should probably only be a kwarg for enum, str, and char channels
* `alarm_status` makes sense (as notes in docstring, it's for sharing one alarm between channels)
* `timestamp`, `status`, `severity` should be lumped into `metadata`, same as in the `set_dbr_data` method
* If alarm-related metadata is provided (i.e., status, severity) it should get applied to the alarm status instance, not silently ignored.

danielballan:
Superceded by #96

PR #92, opened 2017-09-20
MNT: hide import from _data to support py3.4

tacaswell:
attn @dchabot

### DIFF
--- a/caproto/__init__.py
+++ b/caproto/__init__.py
@@ -6,7 +6,10 @@ from ._state import *
 from ._commands import *
 from ._dbr import *
 from ._status import *
-from ._data import *
+try:
+    from ._data import *
+except SyntaxError:
+    pass
 
 from ._version import get_versions
 __version__ = get_versions()['version']


danielballan:
Cool. So we'll try to keep the parts of the code that don't use concurrency (the core, the threading client) 3.4-compatible.

tacaswell:
As part of the test refactor you briefly mentioned, we probably should segregate the 3.4 compatible tests and run them (I think the PV tests copied from pyepics should all be 3.4 compatible).

klauer:
Why are we caring about 3.4?

tacaswell:
Because @dchabot asked me if it worked with python <3.5 and the threaded client does work with 3.4 with only this change.

klauer:
OK, in that case it's fine with me. Can you check for 3.4 in the exception block just in case someone dumb (like me) makes an actual syntax error in `_data.py`?

danielballan:
I'm :+1: on support 3.4, but I don't like conditional top-level imports. Let's take the opportunity to move the `_data.py` content in `caproto/server/core.py`. These are helper utilities for implementing servers, not part of the protocol.

I also plan to reorganize the test modules (we have test_threading_pv, test_threading, test_thread_pv...) which is a good excuse to separate 3.4-compatible tests.

PR #94, opened 2017-09-23
CI: build epics from source; cache on travis

klauer:
Benefits: full matrix of environments to test with (Python versions + EPICS base/module versions)

First run is slow until cache is built (~7 minutes), subsequent builds are on the same order as the current docker solution. See build that uses cache: https://travis-ci.org/klauer/caproto/builds/279039068

~Note that 3.16 is still a bit finicky, but otherwise tests are passing. Will try to investigate it, though I don't think it's too big of a deal at the moment.~

Update on R3.16.1 and 3.16 branch master, pyepics test IOC segfaults consistently when `simulator.py` is running (also from the test IOC). Should report upstream and not bother with 3.16 for now.


danielballan:
Sounds good. Travis is failing. I'm OK with switching to Circle-CI if we can figure out how to get the GH integration working.

klauer:
It's failing because of failing tests... I'd rather not switch to circle after spending a day getting this working...

danielballan:
Oh, of course.

It's great to be testing against a matrix of EPICS versions. Thanks for taking this on.

danielballan:
The failure was an intermittent one involving the threaded client. All green now.

PR #95, opened 2017-09-24
Refactor of ChannelData server utilities

danielballan:
This is in a messy state (tests aren't even close to passing) but posted for discussion.

Goals:
* [ ] Sync channels and alarms so that updates to either propagate correctly.
* [ ] Store the 'ground truth' copy of current data in a graphical DBR instead of in Python attributes. Expose properties for user-friendly access on demand.
* [ ] Handle access permissions
* [ ] Only allow updates through a function call (`write(...0`) which can handle coordinated changes, unlike updates through `setattr`.


danielballan:
Superceded by #96.

PR #96, opened 2017-09-25
Refactoring _data.py (revisited)

danielballan:
A slightly less ambitious take on #95, re-done from scratch with cleaner commits.

- [x] Implement access rights
- [x] One alarm has many channels
- [x] Sync an alarm and all its channels properly
- [x] Only allow updates via the `write` method; make attributes into read-only properties.
- [x] Update metadata when it is included in the write. (It is presently ignored.)

Supercedes #95 and and #90


caproto/_data.py
```diff
@@ -10,6 +10,10 @@
                    SubscriptionType)
 
 
+class Forbidden(Exception):
```
klauer: We have a caproto base exception

caproto/_data.py
```diff
@@ -125,34 +129,50 @@ def convert_values(values, from_dtype, to_dtype, *, string_encoding='latin-1',
     return np.asarray(values).astype(_numpy_map[to_dtype])
 
 
-class ChannelAlarmStatus:
-    def __init__(self, *, channel_data=None, status=0, severity=0,
+class ChannelAlarm:
```
klauer: :+1:

caproto/_data.py
```diff
@@ -125,34 +129,50 @@ def convert_values(values, from_dtype, to_dtype, *, string_encoding='latin-1',
...
-                 alarm_string=''):
-        self.channel_data = channel_data
-        self.status = status
-        self.severity = severity
-        self.acknowledge_transient = acknowledge_transient
-        self.acknowledge_severity = acknowledge_severity
-        self.alarm_string = alarm_string
+                 alarm_string='', string_encoding='latin-1'):
+        self._channels = []
+        self.string_encoding = 'latin-1'
```
klauer: Should be string_encoding

caproto/_data.py
```diff
@@ -219,60 +246,73 @@ def __init__(self, *, value=None, timestamp=None, status=0, severity=0,
...
-    def astype(self, to_dtype):
-        '''Convenience function: convert stored data to a specific type'''
-        native_to = native_type(to_dtype)
-        return convert_values(values=self.value, from_dtype=self.data_type,
-                              to_dtype=native_to,
-                              string_encoding=self.string_encoding,
-                              enum_strings=getattr(self, 'enum_strings', None))
-
-    async def get_dbr_data(self, type_):
+    async def read(self, hostname, username, data_type):
```
klauer: I like it but also feel that Python should be able to get a DBR type out without having to fake a host name and username

caproto/_data.py
```diff
@@ -219,60 +246,73 @@ def __init__(self, *, value=None, timestamp=None, status=0, severity=0,
...
             # TODO: non-standard type request. frequent ones probably should be
             # cached?
-            dbr_metadata = DBR_TYPES[type_]()
+            dbr_metadata = DBR_TYPES[data_type]()
 
         self._copy_metadata_to_dbr(dbr_metadata)
         return dbr_metadata, values
 
-    async def set_dbr_data(self, data, data_type, metadata):
+    async def write(self, hostname, username, data, data_type, metadata):
```
klauer: Ditto - should be able to shove in data via python too without caring about CA access rights

caproto/_data.py
```diff
@@ -331,30 +371,39 @@ def status(self):
...
+        username : string
+        
+        Returns
+        -------
+        access : :data:`AccessRights.READ_WRITE`
+        """
         return AccessRights.READ_WRITE
 
+    def __del__(self):
+        self.alarm.disconnect(self)
```
klauer: Weakref in alarms instead?

caproto/_data.py
danielballan: How about `auth_write` for authenticated writes by the client and `write` for unauthenticated direct access?

caproto/curio/client.py
```diff
@@ -35,10 +35,11 @@ def __init__(self, circuit):
         self.new_command_condition = curio.Condition()
         self._socket_lock = curio.Lock()
 
-    async def create_connection(self):
-        self.socket = await socket.create_connection(self.circuit.address)
-        await self._socket_lock.release()
+    async def connect(self):
```
danielballan: The original version of this is pretty clever but possibly buggy. Its advantage: returning a Channel object before the socket was finished opening. I retreated to a blunter approach for simplicity, but ultimate we might want to restore the original one before merging.

caproto/curio/client.py
```diff
@@ -182,8 +183,9 @@ def process_subscription(self, event_add_command):
         await self.circuit.send(command)
 
         async def _queue_loop():
-            command = await queue.get()
-            self.process_subscription(command)
+            while True:
```
danielballan: :sob:

caproto/curio/client.py
klauer: If I recall correctly, I think it's because I ran into the case where something else had locked the socket and `connect()` was unable to do its thing.

danielballan: I think we should merge this change, actually. The existing code can error out if the `socket` isn't ready by the time the `_receive_loop` task is spawned.The original goal of the complexity here was to make channel creation as asynchronous as possible. We can revisit that goal when we do performance tuning.

danielballan:
There is some cleanup left to do (particularly, masks). But this PR is now at the scope I wanted to reach.

danielballan:
Merging this amalgam PR based a Hangout conversation with @klauer and @tacaswell.

PR #99, opened 2017-10-02
TST: Fix race condition in threading tests.

danielballan:
I think this should fix some intermittent failures we have seen on Travis.

### DIFF
--- a/caproto/threading/client.py
+++ b/caproto/threading/client.py
@@ -12,36 +12,12 @@ from collections import Iterable
 
 import caproto as ca
 
-from typing import Callable, Optional
-
-
-class ConditionType:
-    ...
-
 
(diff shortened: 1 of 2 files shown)


tacaswell:
It is true, those failures are no longer intermittent.

klauer:
Good find - those have been bugging me for a while!

danielballan:
I should add some tests that trip all these timeout errors....

danielballan:
Hmm. This does not fix the race condition. It does simplify the code and give better TimeoutErrors though, so worth keeping.

danielballan:
@klauer called the critical fix correctly, in 6b06969. The other commit is not related to the fix, but I do think it improves the code.

klauer:
A+ for persistence

PR #102, opened 2017-10-05
REF: dbr redux

klauer:
Needs some test fixes that used DBR internals (timestamp-related)

Arguably cleaner - leaving it up to @danielballan and @tacaswell to decide if it's worth it. 
If so, then we can clean `caproto._data` a bit more.


caproto/_dbr.py
```diff
@@ -68,6 +59,99 @@ class AlarmStatus(IntEnum):
...
+    status string, enum string ...) occurs.
+    '''
+
+    DBE_VALUE = 1
+    DBE_LOG = 2
+    DBE_ALARM = 4
+    DBE_PROPERTY = 8
+
+
+ChType = ChannelType
```
danielballan: Maybe we should just pick one name.

caproto/_dbr.py
```diff
@@ -93,923 +217,453 @@ def timestamp_to_epics(ts):
...
+                      'upper_alarm_limit', 'upper_warning_limit',
+                      'lower_warning_limit', 'lower_alarm_limit')
+    control_fields = ('upper_ctrl_limit', 'lower_ctrl_limit')
+    info_fields = ('status', 'severity', ) + graphic_fields
+    _fields_ = [('status', short_t),
+                ('severity', short_t)
+                ]
 
+    @classmethod
+    def build_control_fields(cls, type_):
```
danielballan: Use `data_type` here for consistency.

caproto/_dbr.py
```diff
@@ -93,923 +217,453 @@ def timestamp_to_epics(ts):
...
-    #     ushort_t    value;            /* current value */
-    # };
-    DBR_ID = 10
-    _pack_ = 1
-    _fields_ = [
-        ('status', short_t),
-        ('severity', short_t),
-    ]
+    @classmethod
+    def build_graphic_fields(cls, type_):
```
danielballan: and here

caproto/_dbr.py
```diff
@@ -93,923 +217,453 @@ def timestamp_to_epics(ts):
...
-        ('status', short_t),
-        ('severity', short_t),
-        ('secondsSinceEpoch', ctypes.c_uint32),
-        ('nanoSeconds', ctypes.c_uint32),
-        ('RISC_pad0', short_t),
-        ('RISC_pad1', char_t),
-    ]
+# Native value types
+class DBR_STRING(DbrNativeValueType):
+    DBR_ID = ChannelType.STRING
```
danielballan: :+1: for making this an IntEnum instead of an int

caproto/_dbr.py
klauer: Done, leaving as `ChannelType`

danielballan:
I'm sold. The fact that the original, "brute force" implementation lives on in tests and is proved identical to the new implementation is a big help.

PR #106, opened 2017-10-06
FIX: alarm acknowledgement logic

klauer:
Functionality based on what's in EPICS base, as of yet untested plane commits. Some tests may need to be modified to reflect the logic built into the alarm (specifically transient/severity acknowledge)


klauer:
I think the test failures are _just_ unrelated threaded PV stuff.  Should be ready for review.

PR #110, opened 2017-10-30
Mnt ophyd fixes

tacaswell:
Changes found while trying to get ophyd + caproto to work.  I expect to push more here.


caproto/_commands.py
```diff
@@ -563,10 +563,11 @@ class SearchRequest(Message):
...
     def __init__(self, name, cid, version):
         size, payload = padded_string_payload(name)
-        _len = len(name)
-        if _len > 40:
-            raise CaprotoValueError("EPICS imposes a 40-character limit on "
-                                    "strings. The " "string {!r} is {} "
+        rec, _, field = name.partition('.')
+        _len = len(rec)
+        if _len > 60:
+            raise CaprotoValueError("EPICS 3.14 imposes a 60-character limit "
```
danielballan: Have you encountered any documentation about how different versions of EPICS handle this? The version hardcoded here (3.14) may not match the actual `version` that we know here.

caproto/threading/client.py
```diff
@@ -833,7 +833,12 @@ def get(self, *, count=None, as_string=False, as_numpy=True,
             return info['char_value']
         elif as_string and self.typefull in ca.enum_types:
             enum_strs = self.enum_strs
-            ret = [enum_strs[r] for r in info['value']]
+            ret = []
```
danielballan: Can you add a comment on what we are silently ignoring here?

caproto/_commands.py
tacaswell: See the link in the commit message http://www.aps.anl.gov/epics/tech-talk/2005/msg00083.php

klauer: I think we can safely ignore anything below 3.14.

caproto/threading/client.py
klauer: It's not impossible to have invalid enum values that don't correspond to any string (I think that's what @tacaswell ran into here)

klauer:
Is this still not ready @tacaswell?

caproto/_commands.py
danielballan: Fair enough. I like the idea that if there are any Reagan-era Star Wars satellites out there, we could talk to them with Python 3.6. But until we find one to test against.... ;-)

tacaswell:
I have put 0 cycles into this since pushing.  This is just to record the hacking I did before giving up on the effort.

danielballan:
Rebased to get the test fixes in #115

danielballan:
There are more fixes to come until ophyd + caproto passes the ophyd + pyepics tests, but we might as well merge them incrementally since the work has stalled.

PR #114, opened 2017-11-13
Add CONTRIBUTING instructions for setting up a dev environment

danielballan:


danielballan:
I delete my `.env` stuff in favor of a less elegant but more straightforward approach already merged as part of #115. This PR is now just a docs PR, recording my process for getting the tests running in a fresh Ubuntu VM.

danielballan:
Are we happy with this?

PR #120, opened 2017-11-17
FIX: Standardize string encoding and expose ENV var.

danielballan:
- The UTF-8 encoding is not an appropriate default for EPICS (and often blows up on real cases).
- Now the encoding is configurable through an ENV var.

### DIFF
--- a/caproto/curio/server.py
+++ b/caproto/curio/server.py
@@ -1,6 +1,7 @@
-import logging
-import random
 from collections import defaultdict, deque, namedtuple
+import logging
+import os
+import random
 
 import curio
 from curio import socket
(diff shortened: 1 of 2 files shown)


tacaswell:
What makes utf-8 blow up?!

danielballan:
something something `\x88 is an illegal start byte`

danielballan:
If I understand correctly, all valid ascii is valid UTF-8, but some latin-1 is not valid UTF-8: https://stackoverflow.com/a/7048780/1221924

klauer:
Thanks for fixing this up. 👍 The reality is EPICS is effectively not working with any specific string encoding. A "proper" implementation would always just return byte strings, but this is not so user-friendly to deal with. My opinion is that either 'ascii' or 'latin-1' are about the closest things to what EPICS intended.

PR #121, opened 2017-11-18
REF: move create_many_channels to curio client

klauer:
Need to think about timeouts (search failures, connection failures, slow responses, etc.)


klauer:
I'm incapable of single-issue PRs. Redoing this and lumping with the trio branch for now. 

Can cherry-pick afterwards if trio is not deemed worthy.

PR #122, opened 2017-11-21
Trio client/server test

klauer:
Implements an EPICS Channel Access client and server using [python-trio](https://github.com/python-trio/trio).

This was a lot less work than it looks like, as it's largely a copy-paste of the curio client/server.

Notes:
* Patch in `sendmsg` in trio as it does not yet support it (maybe consider a PR there?)
* Parameterize server tests for doing trio/curio tests easily
* Tests are actually mostly passing - frail benchmarks hang sometimes on on pytest/travis
* - [x] Made @danielballan 's life miserable with ugly, ugly code everywhere
* Future: cancel points, nursery usage should be thought about more carefully
* Future: add into the get/put/etc benchmarks, somehow (help wanted here)

Now in separate PRs:
* Support EPICS_CAS_INTF_ADDR_LIST in adf52d9 for both server implementations
* Refactor `create_many_channels` in curio (still no performance gains?)


danielballan:
<3 the issue label

klauer:
Rebased and working locally; tests aren't so stable on travis. Should probably wait to merge.

danielballan:
To flag a comment from an offline conversation, trio is looking better and better and seems to be building momentum/maturity. I could see shifting from curio to trio as our primary async implementation.

klauer:
Rebased on #178, mostly passing locally.

Note from @danielballan indicates that there's been a rethink about nurserys upstream in trio, and we should still probably wait again to merge this based on that...

PR #128, opened 2017-12-14
FIX: pass through args, kwargs in top-level get_pv

ZLLentz:
I believe this was a typo, this is the only part in the `threading.get_pv` chain that doesn't pass the `*args` and `**kwargs` down to the `PV` object. These should be passed through to match the `pyepics` interface.


klauer:
Good catch, @ZLLentz. Looks like this exposes some bugs in our usage of `get_pv`, or at least indicates there's some missing functionality in there (@tacaswell?)

klauer:
Post-threading client refactor, we do now pass in kwargs to PV initializer (phew)... Closing.

PR #131, opened 2018-01-13
WIP/ENH: High level server idea

klauer:
* Adds `pvproperty`, a property-like way of defining a caproto IOC's structure
* Adds `PVGroupBase` which allow for grouping of `pvproperty` and other `PVGroupBase`s
* Adds `SubGroup`, the class to use when nesting `PVGroup`s

There are 3 simple examples there, which should be the target of your criticism/suggestions more so than the server source itself. I think it's relatively clean. Let me know what you think, in any case!

Things that might not be obvious:
* If you don't specify a `getter` or a `putter`-handler in your `pvproperty`, the group itself handles the event in its `group_read` and `group_write` functions, respectively
* In all the get/put handlers, `self` refers to the group, as you'd expect, whereas `instance` refers to the `ChannelData` instance
* Prefixes are similar to how ophyd does it - but with the option to override the automatic behavior, and also allow for macro expansion of names
* Part of the intention of grouping PVs in the server like this is for future V4 compatibility
* Currency conversion example adds dependency [asks](https://github.com/theelous3/asks) - but this could be just removed altogether

TODO: 
* Add easy way to do `I/O Interrupt`-style PVs, where background events are causing the PV values to update
* Consider an alternate `pvproperty` that defines the `putter` instead of the `getter`, as this will be a common use case
* Alarm groups work by dictionary key in a single group; could potentially allow inheriting of parent group's alarm signal(s).
* A couple prefix/subgroup bugs
* Tests, if everyone likes this enough to proceed

### DIFF
--- a/caproto/curio/high_level_server.py
+++ b/caproto/curio/high_level_server.py
@@ -5,14 +5,13 @@ from types import MethodType
 
 from .. import (ChannelDouble, ChannelInteger, ChannelString,
                 ChannelAlarm)
-from ..benchmarking import set_logging_level
-from .server import start_server
 
 
 logger = logging.getLogger(__name__)
 
 
-class PVSpec(namedtuple('PVSpec', 'get put attr name dtype value alarm_group')):
+class PVSpec(namedtuple('PVSpec',
+                        'get put attr name dtype value alarm_group')):
     'PV information specification'
     __slots__ = ()
     default_dtype = int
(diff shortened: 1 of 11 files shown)


danielballan:
Per discussion in private channels, I am merging this and #136 somewhat prematurely because each makes the other easier to test.

It seems like everyone who has had time to look at this is pretty happy with the shape of it. Tests and bug-fixes will come in future PRs. Note that this PR does not touch any existing files, so it will not break anything.

PR #132, opened 2018-01-14
CI: aps.anl.gov SSL certificate has expired; switch to github links

klauer:
Nuked the travis cache


danielballan:
Good to know NSLS-II isn't the only facility that drops the ball on this from time to time.

PR #133, opened 2018-01-20
MNT: Local dev tweaks

klauer:
Tweak paths for local dev environment

7d21b3f also includes best-yet attempt at ensuring IOCs are running before testing begins


danielballan:
:clap:

PR #140, opened 2018-01-26
Ci fix nightly

tacaswell:


danielballan:
Thanks for this heroic work (in progress...), @tacaswell!

tacaswell:
I really need to report that pip bug upstream....

PR #141, opened 2018-01-27
FIX: Some minor fixes (cli, high-level server)

klauer:


caproto/_cli.py
```diff
@@ -344,7 +344,7 @@ def monitor_cli():
                                  "{timedelta} and usages like "
                                  "{timestamp:%%Y-%%m-%%d %%H:%%M:%%S} are "
                                  "supported."))
-    parser.add_argument('-m', type=str, metavar='MASK',
+    parser.add_argument('-m', type=str, metavar='MASK', default='ma',
```
tacaswell: help string says 'va', this says 'ma'

caproto/_cli.py
```diff
@@ -680,7 +680,7 @@ def put(pv_name, data, *, verbose=False, timeout=1, priority=0, repeater=True):
             for command in commands:
                 if (isinstance(command, ca.WriteNotifyResponse) and
                             command.ioid == req.ioid):
-                    response = command
+                    # response = command
```
tacaswell: why?  Looks like it isn't used?

caproto/_dbr.py
```diff
@@ -20,11 +20,10 @@
     USE_NUMPY = True
 
 
-class AccessRights(IntEnum):
+class AccessRights(IntFlag):
```
tacaswell: nifty! https://docs.python.org/3/library/enum.html#intflag

3.6 is paying divides already.

caproto/_cli.py
klauer: @danielballan ?

danielballan: Relic of a copy/paste I think. Safe to delete.

PR #142, opened 2018-01-27
FIX: Handle a weird socket error in curio server

tacaswell:
- tweak permissions on spoof-beamline
 - handle a weird socket error in curio server


caproto/curio/server.py
```diff
@@ -390,8 +390,14 @@ def __init__(self, host, port, pvdb, *, log_level='ERROR'):
...
             bytes_to_send = self.broadcaster.send(beacon)
+            fail = []
             for addr_port in addresses:
-                await self.udp_sock.sendto(bytes_to_send, addr_port)
+                try:
+                    await self.udp_sock.sendto(bytes_to_send, addr_port)
+                except OSError:
+                    fail.append(addr_port)
+            for adr in fail:
+                addresses.remove(adr)
```
danielballan: What if something goes offline for a bit and then comes back? Should we refresh `get_beacon_address_list()` every N loops and thereby retry any failures?

danielballan: Maybe a more conservative change, to start, with would be to catch the `OSError` but not remove failed addresses from the list.

tacaswell: I apparently had a comment get lost in the ether.

I was seeing this due to something funny in the local networking that failed on the first attempt to send.  This is a "You may not do this thing to this port" error not "the other side did not respond" error.  I don't think it is the sort of thing that would ever get better.

danielballan: Sounds plausible I guess. If @klauer is persuaded by that, this is good to merge.

klauer: Would like to do some more testing on this - maybe `ifdown`/`ifup [interface]` and see what happens?

tacaswell:
```
14:12 $ python scripts/defaultdictserver 
Task 7 crashed
  File "/home/tcaswell/mc3/envs/dd36/lib/python3.6/site-packages/curio/kernel.py", line 826, in _run_coro
    trap = current._send(current.next_value)
  File "/home/tcaswell/mc3/envs/dd36/lib/python3.6/site-packages/curio/task.py", line 96, in _task_runner
    return await coro
  File "/home/tcaswell/src/bnl/epics/caproto/caproto/curio/server.py", line 405, in broadcast_beacon_loop
    await self.udp_sock.sendto(bytes_to_send, addr_port)
  File "/home/tcaswell/mc3/envs/dd36/lib/python3.6/site-packages/curio/io.py", line 251, in sendto
    return self._socket.sendto(bytes, flags, address)
OSError: [Errno 22] Invalid argument
Curio server failed: {<class 'OSError'>}
Task Context.broadcast_beacon_loop failed: [Errno 22] Invalid argument
```

tacaswell:
```
14:26 $ ip addr
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host 
       valid_lft forever preferred_lft forever
2: enp0s31f6: <NO-CARRIER,BROADCAST,MULTICAST,UP> mtu 1500 qdisc pfifo_fast state DOWN group default qlen 1000
    link/ether 54:ee:75:d0:50:5f brd ff:ff:ff:ff:ff:ff
3: wlp4s0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc mq state UP group default qlen 1000
    link/ether 14:ab:c5:6c:77:71 brd ff:ff:ff:ff:ff:ff
    inet 10.3.5.120/21 brd 10.3.7.255 scope global dynamic wlp4s0
       valid_lft 2758sec preferred_lft 2758sec
    inet6 fe80::5fb5:ec48:650:842c/64 scope link 
       valid_lft forever preferred_lft forever
4: docker0: <NO-CARRIER,BROADCAST,MULTICAST,UP> mtu 1500 qdisc noqueue state DOWN group default 
    link/ether 02:42:53:01:f6:9a brd ff:ff:ff:ff:ff:ff
    inet 172.17.0.1/16 scope global docker0
       valid_lft forever preferred_lft forever

```

and adding a print in gives

```
('10.3.7.255', 5065)

```

and going back to `_utils`

```
In [144]: list(caproto._utils.get_netifaces_addresses())
Out[144]: 
[('127.0.0.1', '127.0.0.1'),
 ('10.3.5.120', '10.3.7.255'),
 ('172.17.0.1', '172.17.0.1')]

```

klauer: I think netifaces does not always give correct results for broadcast addresses - either that or I don't understand it well enough. You'd think 127.0.0.1/8 would give a bcast address of 127.255.255.255, for example, above. (I ran into this with docker on circle - [my Python 3 solution](https://github.com/NSLS-II/caproto/pull/151/files#diff-1d37e48f9ceff6d8030570cd36286a61R30))

danielballan: A pragmatic, temporary solution could be to permanently drop any addresses that raise an `OSError`, as @tacaswell originally proposed, but issue a warning so that, if this strategy is too blunt, the effect is not so pernicious.

tacaswell: I wonder if this is related to my 'I have IPs on the same local net" issue?

danielballan:
Superceded by #207. Better to blow up and tell the user how to omit the failing interface (by setting env vars) than silently decline to send messages.

PR #144, opened 2018-01-28
ENH: RPC-style functions

klauer:
This sort of syntax should be forward-compatible with V4 as well.

Example:
```sh
(with initial conditions of low, high = 100, 1000)
$ caput -c prefix:get_random:Process 1
Old : prefix:get_random:Process      0
New : prefix:get_random:Process      1
$ caput -c prefix:get_random:low 1500
Old : prefix:get_random:low          100
New : prefix:get_random:low          1500
$ caput -c prefix:get_random:Process 1
Old : prefix:get_random:Process      1
Error occured writing data.
```

Separate monitor shows:
```sh
$ camonitor prefix:get_random:Status prefix:get_random:Retval
prefix:get_random:Retval       2018-01-28 10:23:57.184804 454
prefix:get_random:Status       2018-01-28 10:23:57.184913 Success

prefix:get_random:Status       2018-01-28 10:24:41.410854 ValueError: empty range for randrange()
```


danielballan:
OK, it took me a couple tries but I get it. Very cool. Do the function arguments with the name `x` always pick up the name of `{parent}{attr_separator}x`, or that separately configurable? Do we have to worry about the case where the desired substring is not a valid Python identifier, such as `{parent}:1`, which would seem to require a function with an argument named `1`?

klauer:
@danielballan the only thing that's fixed are the variable -> pv name mappings, currently. Since the `PVFunction` itself is a `SubGroup`, you can change its prefix and separator just as with the regular `SubGroup` init keyword args. If users really wanted it, I suppose we could offer some mapping (maybe it's premature to ask, but any opinion on the naming stuff @teddyrendahl?)

PR #145, opened 2018-01-30
PERF: Apply mask more efficiently. (Also adds semi-related tests.)

danielballan:


klauer:
Looks reasonable to me. Ready to merge?

PR #146, opened 2018-02-01
CI: use prepackaged EPICS binaries

klauer:
EPICS build dependencies from separate [repository](https://github.com/klauer/epics-on-travis); pre-built on travis without requiring its caching mechanism.

* Downloads/extracts and re-runs through build process as a fallback (CI scripts for building should probably remain here, even if just for reference...)
* Archive size significantly reduced using INSTALL_LOCATION and not caching/releasing source
* Build times should be much more consistent and less reliant on travis cache
* Up for suggestions on reducing ugliness of the implementation (largely filenames I guess?)
~* TODO: Should move over more CI fixes and cleanup from epics-on-travis repo~
* Note: repository forked for NSLS-II, too, https://github.com/NSLS-II/epics-on-travis


danielballan:
Nice approach.

> CI scripts for building should probably remain here, even if just for reference...

On the surface I agree, but I do worry about diverging recipes. Should it be _submoduled_ (play ominous music)? Linked? This would obviate moving over the cleanup.

klauer:
Yeah, @danielballan, submodules for support libs may not be such a bad idea.

klauer:
Passing using the CI scripts in a submodule. Should also be able to use those scripts to build custom IOCs or dependencies per-project, though that's hardly necessary for caproto/ophyd/etc.

.travis.yml
```diff
@@ -30,25 +32,19 @@ addons:
...
+
   - git fetch --unshallow
-  - source .ci/epics-config.sh
-  - bash .ci/install-epics-base.sh
-  - bash .ci/install-epics-modules.sh
-  - bash .ci/install-epics-iocs.sh
-  - bash .ci/run-epics-iocs.sh
+  - export CI_SCRIPTS=${TRAVIS_BUILD_DIR}/.ci/ci-scripts
+  - source ${CI_SCRIPTS}/epics-config.sh
+  - bash .ci/install-from-release.sh "https://github.com/klauer/epics-on-travis/releases/download/${PACKAGED_DEPS}"
```
danielballan: Is this supposed to be `NSLS-II` now, not `klauer`?

klauer: I didn't get the releases on NSLS-II just yet - still need to setup the API key and re-trigger a build on travis.

.travis.yml
```diff
@@ -21,7 +23,7 @@ addons:
   apt:
     packages:
       - graphviz  # for building docs
-      - libreadline6-dev  # epics build deps
+      - libreadline6-dev  # epics deps
```
tacaswell: There is way to patch the build configuration to not need readline (see https://github.com/NSLS-II/lightsource2-recipes/blob/master/recipes-tag/epics-base/config_readline.diff )  Might be worth doing?

klauer: @tacaswell I'd happily accept a PR over at klauer/epics-on-travis!

danielballan: I'm opened an issue there and tagged Tom. :-D

danielballan:
It worked great on master. Pretty good build times. Thanks for doing this, @klauer! Looking forward to trying out the local dev script later today.

danielballan:
I expect it's much slicker than mine...

PR #150, opened 2018-02-07
API/REF: Standardize examples.

danielballan:
- Move examples that are in doc/source/examples into package.
- *Shorten* some names. (Yes, you read that right.)
- Make a prefix command-line argument optional.
- Consolidate `main()` into `__main__` to simplify things.


caproto/curio/server.py
```diff
@@ -32,7 +32,6 @@ class AsyncLibraryLayer:
 
 class CurioAsyncLayer(AsyncLibraryLayer):
     name = 'curio'
-    sleep = curio.sleep
```
tacaswell: no naps?

tacaswell: ah, I see, expect the library to provide `sleep`

klauer: Fix here could be `sleep = staticmethod(curio.sleep)` if we think it's desirable to increase the scope of this 'async library layer'

tacaswell: no, I think we should only shim out when we have to and I strongly suspect everyone has 'sleep'.

PR #151, opened 2018-02-08
CI: update circleci configuration

klauer:
This is in a much better spot now, but there's still a lot left to do on the time sink that is CI.

* The tests are running natively in a docker image I created which contains EPICS 3.14, 3.15, 3.16, motor, areaDetector, and PVA libraries (using [epics-on-travis](https://github.com/klauer/epics-on-travis) and [caproto-docker](https://github.com/klauer/caproto-docker)).
* IOCs run on tmux in the background when started up
* As-is, only one configuration is actually tested (no conda, so 3.6 only)

**TODO**
- [ ] Some threading tests are failing
- [ ] Some benchmarks are failing (no output after 1 minute kills the build)
- [ ] Install conda on the docker image
- [ ] Get matrixed builds working somehow

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -29,6 +29,8 @@ jobs:
             python -m venv caproto-venv
             . caproto-venv/bin/activate
             pip install -r test-requirements.txt
+            # add pytest-timeout for test failure debugging purposes
+            pip install pytest-timeout
             python setup.py install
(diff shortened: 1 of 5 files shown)


klauer:
This is good enough for now, though passing is very intermittent; self-merging.

PR #154, opened 2018-02-14
FIX/ENH: High-level server tweaks + typhon example

klauer:
Merge #155  before this one and rebase

* Start of smoke test for coverage of IOC examples
* Includes caproto -> ophyd -> pydm/typhon example titled "integration"


danielballan:
Very nice system for testing the IOCs. Thanks for fixing my docs -- I should have covered that in #150.

PR #157, opened 2018-03-07
ENH/REF: high-level server (areaDetector image example, record fields, read-only properties)

klauer:
## Major points
* Adds `ophyd.Device` -> `caproto.PVGroup` "converter" with example
    - Components default to 'unknown' dtype, but capable of getting dtype
      through channel access if a server is running
* Relying on recordwhat, adds ability to mock records with a PVGroup
  consisting of known V3 fields
    - Adds converter from recordwhat record information to PVGroup
    - Does not litter top-level pvdb with all fields, instead exposes field access API at the server-level
* Adds lighter-weight SubGroups, specifiable through dict entries.
  Functionality can then be specified through, for example,`@group.attr.putter` and `@group.attr.startup`
* Adds pvproperty_with_rbv, implemented with the above dict specification of a
  subgroup. Example of this and the above:
```python
pvproperty_with_rbv = get_pv_pair_wrapper(setpoint_suffix='',
                                          readback_suffix='_RBV')
class Group(PVGroup):
    pair = pvproperty_with_rbv(dtype=int, doc='pair1')

    @pair.setpoint.putter
    async def pair(self, instance, value):
        ...
```

## Minor points
* Rename PVGroupBase -> PVGroup
* Moves previous ophyd.Device->caproto.PVGroup conversion tool to a separate
  module
* Placeholder for 3.15 filter API, with new field access API in the server
* Fixed prefixes of sub-subgroups
* Fixed `pvfunction.__call__`, and implemented on top of new dict-based SubGroups
* Fixed IPython-introspection-related bug

## TODO
- [ ] Think about immutability of values at the level of PVGroups for simultaneous access in V4 (future TODO)
- [ ] In follow-up PR, re-arrange all non-curio modules to be in a separate submodule (high-level server stuff, records, conversion, etc.)
- [x] Tests and coverage
- [x] Flesh out a `RecordField` `PVGroup` to mock a record
- [x] Use `mock_record` when the given record type is available in the ophyd->caproto conversion routine (... added to issue list to be done later)
- [x] self -> obj in case of wrapped method not referring to top-level PVGroup
- [x] Support `__iter__` in `Context`, allowing `__getitem__` to remain


caproto/curio/server.py
```diff
@@ -312,42 +312,98 @@ def __init__(self, host, port, pvdb, *, log_level='ERROR'):
                 commands = self.broadcaster.recv(bytes_received, address)
                 await self.command_bundle_queue.put((address, commands))
 
-    async def broadcaster_queue_loop(self):
+    def __getitem__(self, pvname):
```
tacaswell: Defining `__getitem__` also provides an implicit `__iter__`, maybe use a normal method there?

tacaswell: or support more of the dict interface.

danielballan:
Rename `self` when it's not `self`. Maybe `obj`? Just to flag that the decorator is affecting what is passed in.

caproto/ioc_examples/setpoint_rbv_pair.py
```diff
@@ -18,22 +18,24 @@ class Group(PVGroup):
     # Creates {prefix}pair2 and {prefix}pair2_RBV
     pair2 = pvproperty_with_rbv(dtype=float, doc='This is pair2')
 
+    # TODO: @danielballan to rename the above appropriately
```
danielballan: @danielballan will strive to live up to this responsibility.

PR #159, opened 2018-03-11
WIP: asyncio repeater fix

klauer:
Current implementation never closes UDP sockets and never detects them as having errored out.

PR implementation of `error_received` at least removes them from `remotes`, but according to `lsof` the sockets still exist. To investigate further.


danielballan:
Decided to just remove this code for now.

PR #164, opened 2018-03-16
CLN: Remove 'sync.simple_client', superceded by CLI code.

danielballan:
This is very old code, a single-shot client that gets, puts, monitors, and exits. Its purpose was as a test and simple reference client implementation. The code in `caproto/_cli.py` supercedes it. No need to keep maintaining this.


caproto/tests/test_examples.py
```diff
@@ -384,6 +359,47 @@ def find_put_value(pv):
         print(f'Read {pv} = {value}')
 
 
+@pytest.mark.parametrize(
+    'module_name, pvdb_class_name, class_kwargs',
+     [('caproto.ioc_examples.currency_conversion', 'CurrencyConversionIOC', {}),
```
tacaswell: You need to remove this one from here as it is run below

PR #166, opened 2018-03-16
FIX: Don't throw out the baby with the bathwater.

### LINKED ISSUE #163: Allow mix of Registration message and search (etc) in synchronous broadcaster
As implemented currently, the RegistrationRequest is intercepted and not broadcast, but any subsequent messages are similarly not broadcast.

Fix by removing RegistrationRequest from message list and broadcasting remaining messages to clients. E.g., `[Version(), RegistrationRequest(), SearchRequest()]` -> intercept registration, rebroadcast `[Version(), SearchRequest()]`

Reference https://github.com/epics-base/epics-base/blob/3.15/src/ca/client/repeater.cpp#L555

danielballan:
Closes #163


caproto/sync/repeater.py
```diff
@@ -127,20 +136,19 @@ def _run_repeater(server_sock, bind_addr):
...
-                if other_port != port:
-                    try:
-                        server_sock.sendto(bytes_to_broadcast, (other_host,
-                                                                other_port))
-                    except Exception as ex:
-                        to_remove.append((other_host, other_port))
-
-            if to_remove:
-                _update_all(clients, servers, remove_clients=to_remove)
+                commands.remove(command)
```
tacaswell: I have a slight preference for building a new list/deque of commands that should be set out rather than mutating the input.  I am not 100% sure on the equality semantics of commands and worry this might introduce weird corner cases.  Also this is a linear search + a bucket load of shifts,  suspect build a new dequeue would be faster.

Could also pre-allocate a list of the maximum length full on `None`, fill it in and then slice to only what you need below.

*end premature optomization*

PR #170, opened 2018-03-16
DOC: Update minimum verison of Python.

danielballan:


tacaswell:
Are we using the right classfiers in setup.py?

danielballan:
I have never paid attention to how to do generate those. (I don't think any of our projects do.) But we should!

Issue #153, opened 2018-02-10
Rework threading client


danielballan:
Notes to self. Implementation in progress.

1. The user asks for one or more PVs, providing name and priority.
2. The Context:
    * Have I already got a `PV` instance for this? If so, return that. If not, allocate a `cid` and instantiate a corresponding `PV` object.
    * For all the _new_ `PV` objects just created above, send `[(name, cid), (name2, cid2)]` to the Broadcaster to search.
     * Cache the `cid -> PV` mapping and return a list of `PV` objects, some or all of which may not be connected yet.
3. The Broadcaster:
    * Do I have an unexpired search result for this? If so, send `(address, (cid1, cid2, ...))` to the `Context`. If not, send a `SearchRequest` using the `cid` provided by the `Context`.
    * In the recv loop, send `(address, (cid1, cid2, ...))` to the `Context` as updates come in. (I think that _all_ Contexts are current notified, which is a problem because they have different cids.... this could be solved using a queue per context, as we do in `ChannelData`.)
4. The Context:
    * Receives `(address, (cid1, cid2, ...))` from the Broadcaster in its channel creation queue.
    * Looks up the `cid` in its cache to find the `PV`, which knows the `priority`. Checks: do I have a `VirtualCircuitManager` for this `(address, prority)` yet? If not, create a socket and a `VirtualCircuitManager`.
    * Creates a `caproto.ClientChanel`. (There is not need for a special `Channel` class.)
    * Tacks a reference to this `caproto.ClientChannel` and to the `VirtualCircuitManager` onto the `PV`.
    * Sends creation messages via `VirualCircuitManager.send`.
5. The VirtualCircuitManager:
     * Receives channel creation confirmation.
     * Looks up `cid` on the `VirtualCircuitManager`. Tells the `PV` it is `connected` via updating a `connected` attribute and possibly setting an `Event`, to make it possible to wait on `PV` connection.

danielballan:
Closed by #161

PR #173, opened 2018-03-23
WIP: fixing caproto integration with ophyd

mrakitin:
An attempt with @tacaswell to fix integration of ophyd & caproto. See https://github.com/NSLS-II/ophyd/pull/523 as well.


danielballan:
Thanks, @mrakitin! It's fun to see you here on caproto.

We believe the important bits of this were incorporated into #153 and merged into master. But there is still at least one unknown issue making a handful of ophyd tests fail with caproto, so let's leave this PR open in case the answer is in here somewhere.

danielballan:
As noted above, #153 picked off the important bits of this. #201 tied up the loose ends. Closing!

PR #183, opened 2018-04-14
REF/TST: refactor and test type conversions, switchable array/numpy backends

klauer:
Supersedes #191 

This PR represents a big change in data type conversions. While far, _far_ from pretty, the functionality appears to more closely match that of epics-base (compared through `ca_test`), which I made some effort to add in `test_type_conversions.py`.

Additionally, this adds backend support through `caproto._backend`. The default is numpy if available, otherwise it falls back to `array.array`. There is a controversial subclassing of `array.array` in the array backend, as I felt it was the only safe way to implement it. We can rework this at a later date.

TODO
- [x] check test/conversion coverage
- [x] array.array / numpy backend switch (or parametrize for testing)
- [x] add some up/downcast conversion tests (int/long/float/double) (noted at end of tests)
- [x] mostly checked array.array compatibility
- [x] string array conversion tests
- [ ] eventually make conversions readable/etc
- [ ] reconsider (im)mutability in data classes, despite performance implications


danielballan:
Perhaps we could borrow ideas from or vendor or even depend on (!) https://github.com/mrocklin/multipledispatch for this.

klauer:
https://github.com/epics-base/epics-base/blob/3.14/src/db/dbConvert.c  :face_with_thermometer:

klauer:
OK, current status is:
```
Failure Count / Tests
---------------------
7 caproto/tests/test_examples.py::test_typhon_example
1 caproto/tests/test_examples.py::test_curio_client_example
1 caproto/tests/test_server.py::test_with_caget[array-trio-enum-ChannelType.GR_FLOAT]
```
These are all passing locally and on travis for klauer/caproto without the merge required by #190 . Not sure why that is. I think I'm done with this PR for now.

PR #184, opened 2018-04-15
FIX: threading callback queue

klauer:
Fixes NSLS-II/ophyd#524 by avoiding callback deadlock

Solution was the first that came to mind (ophyd pseudo positioner tests all pass locally) - eventually smarter handling using a thread pool or throwing out monitors that are too far backlogged may be worthwhile.

### DIFF
--- a/caproto/threading/client.py
+++ b/caproto/threading/client.py
@@ -745,7 +745,9 @@ class Context:
 class VirtualCircuitManager:
     __slots__ = ('context', 'circuit', 'channels', 'ioids', 'subscriptions',
                  '_user_disconnected', 'new_command_cond', 'socket',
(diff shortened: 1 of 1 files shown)


danielballan:
Nice!

>  eventually smarter handling using a thread pool

Indeed. There is at least one other place we have talked about doing this. I made an issue: https://github.com/NSLS-II/caproto/issues/185

> or throwing out monitors that are too far backlogged may be worthwhile.

Interesting. Servers are specifically allowed to drop responses (sorted by priority). I wonder if the protocol gives any guidance about clients. Either way, this is worth considering.

Big-picture: both of those issues are larger concerns, and this fixes this ophyd tests. Merging!

PR #190, opened 2018-04-20
FIX: protect more things with lock

tacaswell:
Prevents race condition of

 - am I connected?
 - (connection happens, flips event)
 - clear event
 - wait on event forever


danielballan:
The test failures here are random. Merging -- thanks @tacaswell!

PR #191, opened 2018-04-21
REF: add assuredly controversial DbrStringArray

klauer:
A bug in the current handling of strings was highlighted when running:
```python
color_mode = caget('13SIM1:image1:ColorMode_RBV', verbose=True).data[0].decode('ascii')
```

This would fail when the buffer from EPICS had garbage after the first null terminator, because `native_to_builtin` wasn't stripping it away effectively.

Other layers (sync, curio, trio) and especially the user should not have to worry about finding null terminators in DBR_STRING types (single- or multi-string lists). Current master returns `np.array` with a dtype of '<S40' which, while accurate, is rather inconvenient to work with.

That being said, we could just tweak `native_to_builtin` to remove everything after the null terminator, then `ljust` with null terminator chars, using a np.array dtype='<S40' as the current master. Instead of that safe route, I decided to submit a PR that @danielballan and @tacaswell will assuredly turn their noses up at :) .

### DIFF
--- a/caproto/tests/test_serialization.py
+++ b/caproto/tests/test_serialization.py
@@ -162,8 +162,8 @@ payloads = [
 
     (ca.ChannelType.STRING, 1, b'abc'.ljust(40, b'\x00'), None),
     (ca.ChannelType.STRING, 3, 3 * b'abc'.ljust(40, b'\x00'), None),
-    (ca.ChannelType.STRING, 3, numpy.array(['abc', 'def'], '>S40'), None),
-    (ca.ChannelType.STRING, 3, numpy.array(['abc', 'def'], 'S40'), None),
+    (ca.ChannelType.STRING, 2, numpy.array(['abc', 'def'], '>S40'), None),
+    (ca.ChannelType.STRING, 2, numpy.array(['abc', 'def'], 'S40'), None),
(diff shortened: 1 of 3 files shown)


caproto/tests/test_serialization.py
```diff
@@ -162,8 +162,8 @@ def make_channels(cli_circuit, srv_circuit, data_type, data_count, name='a'):
 
     (ca.ChannelType.STRING, 1, b'abc'.ljust(40, b'\x00'), None),
     (ca.ChannelType.STRING, 3, 3 * b'abc'.ljust(40, b'\x00'), None),
-    (ca.ChannelType.STRING, 3, numpy.array(['abc', 'def'], '>S40'), None),
```
tacaswell: How was this passing before?!

tacaswell:
I suggest moving this logic outside of the `NUMPY` block and just always handle strings this way.

There is not much good reason to strings into numpy arrays.

caproto/tests/test_serialization.py
klauer: Not entirely sure...

klauer:
So... not as controversial as imagined?

danielballan:
Closing, superceded by #183

PR #193, opened 2018-04-24
REF: rework sync client search/register a bit

klauer:
This hopefully should make the CLI tests fail a bit less frequently.

In summary:
* multiple search requests (could use improvement)
* registration response not waited for

### DIFF
--- a/caproto/sync/client.py
+++ b/caproto/sync/client.py
@@ -170,8 +170,11 @@ def make_channel(pv_name, logger, udp_sock, priority, timeout):
 
 
 def spawn_repeater(logger):
-    subprocess.Popen(['caproto-repeater', '--quiet'], cwd="/",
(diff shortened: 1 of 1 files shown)


tacaswell:
from the title I expect a +1k/-1k diff....

caproto/sync/client.py
```diff
@@ -60,73 +59,83 @@ def recv(circuit):
...
+                               f"from the search for {pv_name!r}")
+
+    # Initial search attempt
+    send_search()
+
+    # Await a search response, and keep track of registration status
+    retry_timeout = timeout / max((max_retries, 1))
+    t = time.monotonic()
+    retry_at = t + retry_timeout
+    registered = False
```
tacaswell: do we even need to track this anymore?

klauer:
Thanks @tacaswell!

PR #196, opened 2018-05-01
CI/CLN/STY: Add flake8 check to build.

danielballan:
* Add flake8 step to Travis-CI.
* Configure to omit vendored files and allow 115-character line length (!)
* Configure to ignore unused imports in `__init__.py` files.
* Fix all other errors.


jrmlhermitte:
Looks good to me. Flake8 is running as intended with a Travis run on a bad commit [here](https://travis-ci.org/NSLS-II/caproto/jobs/373568578) so looks good. Merging!

PR #197, opened 2018-05-01
DO NOT MERGE: verifying that flake8 fails build on Travis.

danielballan:


danielballan:
Confirmed. Hat tip to @jrmlhermitte for suggesting this diligence.

PR #198, opened 2018-05-01
API: Be more intentional with public API.

### LINKED ISSUE #158: Revisit public API
Now that the client and server implementations are almost feature complete, we should revisit the public API and potentially cull it so that `caproto.<TAB>` is a bit less crowded.

danielballan:
This reduces the number of public objects (``caproto.<TAB>``) from 361 to about 150.

Following h11 and trio, there are three tiers of public-ness here:

1. Things that start with `_` are not intended to be used outside the module where they are defined.
2. Things that do not start with `_` may be imported by other modules in the package but are not
   necessarily public to users.
3. Things that are in `__all__` are fully public and "re-exported" into the `caproto.` namespace.
   Everything that you need to implement a client or server is fully public.

Closes #158


danielballan:
Rebased on #199 to get the critical fix to `special_types`.

tacaswell:
Want to re-base the on master now that #199 is merged?

danielballan:
Yes. Rebased. Let's let the tests spin and also let @klauer take a look at this one before it goes in.

danielballan:
It _seems_ like these failures are random timeouts, but the failure rate is high, particularly of `caproto/tests/test_examples.py::test_curio_server_example`, so I have reservations.

FWIW, locally, that runs and hangs at the end of the test for me -- but it also did that in the last PR (now `master`) for which we got all 8 builds to pass.

tacaswell:
_all_ failing seems like a bad sign!

klauer:
:+1: looks good to me. Probably should wait on #201 now, though...

danielballan:
Rebased with some fixes to adjust for subsequent changes (mainly the big server refactor).

This was previously approved by @tacaswell and given the :+1: by @klauer, so I will self-merge. There were a couple of the usual random failures, but at least one build fully passed.

PR #201, opened 2018-05-02
Locks locks locks

danielballan:
~WIP DO NOT MERGE~

This turned out to be a total rewrite of the threading client with a clearer API, more thoughtful locking, and much more thorough tests.

### DIFF
--- a/caproto/threading/client.py
+++ b/caproto/threading/client.py
@@ -954,7 +954,8 @@ class VirtualCircuitManager:
             elif isinstance(command, (ca.ReadNotifyResponse,
                                       ca.WriteNotifyResponse)):
                 ioid_info = self.ioids.pop(command.ioid)
-                if time.monotonic() > ioid_info['deadline']:
+                deadline = ioid_info['deadline']
+                if deadline is not None and time.monotonic() > deadline:
                     logger.warn(f"ignoring late response with "
                                 f"ioid={command.ioid} because it arrived "
                                 f"{time.monotonic() - ioid_info['deadline']} "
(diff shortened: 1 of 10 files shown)


caproto/threading/client.py
```diff
@@ -893,8 +893,7 @@ def send(self, *commands):
         # be send, and convert them to buffers.
         buffers_to_send = self.circuit.send(*commands)
         # Send bytes over the wire using some caproto utilities.
-        # ca.send_all(buffers_to_send, self._socket_send)
-        self.socket.sendall(b''.join(buffers_to_send))
+        ca.send_all(buffers_to_send, self._socket_send)
```
danielballan: Phew. I read this original change and was like "Whaaaaa? I understand nothing anymore...."

caproto/threading/client.py
```diff
@@ -1282,22 +1284,33 @@ def read(self, *args, timeout=2, **kwargs):
                 f"not respond to attempt to read channel named "
                 f"{self.name!r} within {timeout}-second timeout."
             )
+
+        # TODO shouldn't we get access to the ioid of our request?
```
danielballan: Good point. I think the approach was inherited from pyepics. We can and should be more precise about which reading we return.

caproto/threading/client.py
```diff
@@ -1282,22 +1284,33 @@ def read(self, *args, timeout=2, **kwargs):
...
         "Write a new value and await confirmation from the server."
         # TODO: change use_notify default value; may break tests
         ioid = self.circuit_manager._ioid_counter()
         command = self.channel.write(*args, ioid=ioid,
                                      use_notify=use_notify,
                                      **kwargs)
+        if not use_notify:
+            # TODO:
+            # if wait or callback is not None:
+            #      raise ValueError('Not guaranteed a WriteNotifyResponse')
```
danielballan: Un-commenting this seems correct. There is no `WriteResponse`.](https://github.com/NSLS-II/caproto/blob/master/caproto/_commands.py#L1111) We should not expect any response.

danielballan:
I just caught this title change, @klauer. Well done.

PR #203, opened 2018-05-05
REF: sentinel -> enum

klauer:
A possibility - no hard feelings if not merged


danielballan:
It's fair to question whether the existing system is useful enough to warrant its eccentric implementation. Let me sleep on it.

danielballan:
Haven't forgotten about this. Feeling indecisive about it. Will return to it before 0.1.0.

danielballan:
Updated documentation (which mentioned sentinels being instances of themselves -- not true with this PR) and rebased. Will merge if tests pass.

PR #205, opened 2018-05-09
WIP: Test on Windows with appveyor

danielballan:
- [x] Install and import caproto, following a template provided by trio. [So far, so good.](https://ci.appveyor.com/project/DanielAllan/caproto)
- [ ] Switch all ioc_examples to use trio instead of curio.
- [ ] Skip (on Windows only) all tests involving curio, which seems to not support Windows at all.
- [ ] Get some epics-base stuff running to test against (???) -- or simply skip anything that doesn't test against caproto IOCs.
- [ ] Write a variant of `caproto.threading.client.SelectorThread` for Windows. (The current implementation uses a built-in Python module only available on UNIX.)
- [ ] ~Get appveyor to run on PR branches (notice this branch is on upstream).~ Maybe this works now? Not 100% sure yet.
- [x] Integrate with GitHub to get the box by the merge button.

Possibly of interest to @johnsinsheimer.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,27 @@
+skip_tags: true
+
+os: Visual Studio 2015
+
+environment:
+  matrix:
+    - PYTHON: "C:\\Python36"
+    - PYTHON: "C:\\Python36-x64"
+
+build_script:
+  - "git --no-pager log -n2"
+  - "echo %APPVEYOR_REPO_COMMIT%"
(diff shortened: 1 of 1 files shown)


danielballan:
I'm going to merge this so that we can gradually increase the number of tests run in appveyor. It will never be 100% unless curio comes to support Windows, so there is no sense in waiting.

PR #208, opened 2018-05-16
REF: Use trio in IOC examples.

danielballan:
Addresses one part of #205. ~I get some failures locally but I get the same
failures on the master branch, so I must have a momentary local problem
because Travis is fine on the master branch. We'll see how Travis likes this.~

### DIFF
--- a/caproto/ioc_examples/all_in_one.py
+++ b/caproto/ioc_examples/all_in_one.py
@@ -2,7 +2,7 @@
 import logging
 
 from caproto.benchmarking import set_logging_level
-from caproto.curio.server import start_server
+from caproto.trio.server import start_server
 from caproto.server import (pvproperty, PVGroup)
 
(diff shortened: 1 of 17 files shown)


danielballan:
*A voice calls into the void...*

If anyone happened to have Windows VM with caproto installed, it would be interesting to know whether

```
python run_tests.py caproto/tests/test_examples.py::test_ioc_examples
```

passes on Windows.

danielballan:
#232 will make this a lot easier to do, via the `--async-lib` arg to the IOC examples. And given that @klauer got curio working on Windows I'm not sure what the best default for `--async-lib` is at this point. Might make sense to go with `asyncio` in the spirit of "zero deps".

danielballan:
So, closing in favor of some easier approach on top of #232.

PR #211, opened 2018-05-20
ENH: first pass at a coupled motor + detector simulation IOC

tacaswell:


klauer:
Cool example! :+1:  

As a side note of my preference, I think I prefer:
```python
prop = pvproperty(...)

@prop.putter
def prop(self, instance, value):
    ...
```

to

```python
def prop_put(self, instance, value):
    ...

prop = pvproperty(..., put=prop_putter)
```

That said, a benefit of the high-level server setup is that you can pick and choose between these.

danielballan:
Note: Before this is merged, we should merge in subsequent updates that were made directly to https://github.com/NSLS-II/tutorial/blob/master/demo_ioc.py _except for the `os.environ` block_.

caproto/ioc_examples/jittery_read.py
```diff
@@ -1,4 +1,13 @@
 #!/usr/bin/env python3
+
+import os
+
+# Set some enviornment variables before importing pyepics.
```
danielballan: I think this block should not be added.

tacaswell:
@klauer why?

danielballan:
Tests passed, but the build failed on `flake8`.  I pushed 9bcc41b to fix that and then canceled the Travis build to make room for other stuff.

PR #214, opened 2018-05-23
FIX: Lock unanswered_searches when modifying or iterating over it.

danielballan:
I got a 'dict changed size during iteration' error when using the threading
client in the wild. I have never seen it before or reproduced it since, but I
believe this will fix it.


tacaswell:
do we already lock where we _add_ to this dictionary?

danielballan:
Yes. I searched the file for all references to the dict and verified that.

PR #215, opened 2018-05-25
REF: abstract ioctl call in threaded selector

### LINKED ISSUE #149: Add windows-compatible CLIs
This should be trivial: the UNIX-specific component of the CLI 'scripts' is extremely thin, and so the Windows one should be simple.

### LINKED ISSUE #174: Simple example IOC in Windows error message
I cloned the master branch just now.  Should the simply.py example run properly in Windows yet?
Although I also just realized I don't need to run this in Windows for the application I'm working on...

C:\Users\jsinsheimer\AppData\Local\Continuum\Anaconda3\envs\python=3.6>python "C:\Users\jsinsheimer\AppData\Local\Continuum\Anaconda3\envs\python=3.6\Lib\site-packages\caproto\ioc_examples\simple.py"
PVs: ['simple:A', 'simple:B']
Exception OSError(10022, 'An invalid argument was supplied', None, 10022, None) from selector_select ignored
  File "C:\Users\jsinsheimer\AppData\Local\Continuum\Anaconda3\envs\python=3.6\lib\site-packages\curio\kernel.py", line 739, in _run_coro
    events = selector_select(timeout)
  File "C:\Users\jsinsheimer\AppData\Local\Continuum\Anaconda3\envs\python=3.6\lib\selectors.py", line 323, in select
    r, w, _ = self._select(self._readers, self._writers, [], timeout)
  File "C:\Users\jsinsheimer\AppData\Local\Continuum\Anaconda3\envs\python=3.6\lib\selectors.py", line 314, in _select
    r, w, x = select.select(r, w, w, timeout)

### LINKED ISSUE #210: Add Windows support
An umbrella issue copied from from #205:

- [x] Install and import caproto, following a template provided by trio. [So far, so good.](https://ci.appveyor.com/project/DanielAllan/caproto)
- [ ] Switch all ioc_examples to use trio instead of curio. (#208)
- [ ] Figure out how to put the `scripts/` on the Windows PATH. Seems like the current `setup.py` configuration doesn't do the trick, unclear if we should be using `entry_points` or what. (#215)
- [x] ~Skip (on Windows only) all tests involving curio, which seems to not support Windows at all.~
- [ ] Get some epics-base stuff running to test against (???) -- or simply skip anything that doesn't test against caproto IOCs. (#215)
- [ ] Write a variant of `caproto.threading.client.SelectorThread` for Windows. (The current implementation uses a built-in Python module only available on UNIX.) (#215)
- [ ] ~Get appveyor to run on PR branches (notice this branch is on upstream).~ Maybe this works now? Not 100% sure yet.
- [x] Integrate with GitHub to get the box by the merge button.

Possibly of interest to @johnsinsheimer.

klauer:
aka
# ENH: Add support for Windows.

Closes #210 
Closes #149 
Closes #174


setup.py
```diff
@@ -31,6 +31,16 @@
...
+          'console_scripts': [
+              'caproto-get = caproto.sync.client:get_cli',
+              'caproto-put = caproto.sync.client:put_cli',
+              'caproto-monitor = caproto.sync.client:monitor_cli',
+              'caproto-repeater = caproto.sync.client:repeater_cli',
+              'caproto-defaultdictserver = caproto.examples.defaultdictserver:main',
+              'caproto-spoof-beamline = caproto.examples.spoof_beamline:main',
+          ],
+      },
       scripts=glob.glob('scripts/*'),
```
danielballan: If I am interpreting [the documentation](https://setuptools.readthedocs.io/en/latest/setuptools.html#automatic-script-creation) correctly, these `entry_points` work on POSIX too, so we can delete `scripts=`.

setup.py
```diff
@@ -31,6 +31,16 @@
                 'caproto.threading',
                 'caproto.trio',
                 ],
+      entry_points={
+          'console_scripts': [
+              'caproto-get = caproto.sync.client:get_cli',
+              'caproto-put = caproto.sync.client:put_cli',
+              'caproto-monitor = caproto.sync.client:monitor_cli',
+              'caproto-repeater = caproto.sync.client:repeater_cli',
+              'caproto-defaultdictserver = caproto.examples.defaultdictserver:main',
```
danielballan: :+1: on the renames here.

caproto/_utils.py
```diff
@@ -133,6 +143,10 @@ def get_environment_variables():
             result.get('EPICS_CA_AUTO_ADDR_LIST', '').upper() != 'NO'):
         warn("EPICS_CA_ADDR_LIST is set but will be ignored because "
              "EPICS_CA_AUTO_AUTO_ADDR_LIST is not set to 'no'.")
+        ca_auto_addr_list = result.get('EPICS_CA_AUTO_ADDR_LIST')
+        ca_addr_list = result.get('EPICS_CA_ADDR_LIST')
+        warn(f"EPICS_CA_ADDR_LIST={ca_addr_list!r}")
```
danielballan: The snarky commit message associated with this change is fair. :- )

Can these be moved into the same, single warning though?

caproto/_utils.py
```diff
@@ -602,3 +626,29 @@ def __call__(self):
...
+    '''
+    if available_buffer is None:
+        available_buffer = array.array('i', [0])
+
+    ok = fcntl.ioctl(sock, termios.FIONREAD, available_buffer) >= 0
+    return (max((available_buffer[0], default_buffer_size))
+            if ok else default_buffer_size)
+
+
+if sys.platform == 'win32' or fcntl is None:
```
danielballan: Defining one version of `socket_bytes_available` _or_ the other is probably better than defining the POSIX one and then, maybe, shadowing it. As written, it's impossible to know what (if anything `_windows_compat` will shadow without cross-referencing.

_Not a criticism of this expedient approach, just a bookmark for one of us to consider cleaning this up._

caproto/tests/conftest.py
```diff
@@ -211,7 +214,10 @@ def stop_repeater():
         return
 
     logger.info('[Repeater] Sending Ctrl-C to the repeater')
-    _repeater_process.send_signal(signal.SIGINT)
+    if sys.platform == 'win32':
+        _repeater_process.terminate()
```
danielballan: I think we can simplify by using this method on both platforms. On POSIX it [sends SIGTERM](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.terminate) which is just as good.

caproto/tests/conftest.py
```diff
@@ -56,9 +56,11 @@ def run_example_ioc(module_name, *, request, pv_to_check, args=None,
 
     def stop_ioc():
         if p.poll() is None:
-            logger.debug('Sending Ctrl-C to the example IOC')
-            p.send_signal(signal.SIGINT)
-            logger.debug('Waiting on process...')
+            if sys.platform != 'win32':
+                logger.debug('Sending Ctrl-C to the example IOC')
+                p.send_signal(signal.SIGINT)
```
danielballan: Use `p.terminate()`?

caproto/_utils.py
klauer: It's actually more directed at appveyor - despite setting both, we don't see them in `os.environ`. Should try to confirm that when one of us tests it locally.

caproto/tests/conftest.py
klauer: SIGINT (Ctrl-C) and SIGTERM are different. Perhaps we should be adding SIGTERM handlers to our long-running processes to make them clean up and exit gracefully?

caproto/_utils.py
klauer: I don't disagree it could be written a lot more clearly, but I also somewhat like shoving all the smelly windows support nonsense in its own corner.

caproto/tests/conftest.py
danielballan: Until I just checked, I knew they were different but didn't realize that SIGTERM did not have a default handler in Python to convert it to `KeyboardInterrupt`. I'll open a separate issue about adding SIGTERM handlers, and we'll leave this more expedient solution in place for this PR.

danielballan:
Out of 10 builds:

```
4 caproto/tests/test_examples.py::test_curio_server_example
1 caproto/tests/test_server.py::test_with_caget[numpy-trio-int-ChannelType.ENUM]
1 caproto/tests/test_bench.py::test_many_connections_same_ioc[INFO-trio-connections:{}-100]
1 caproto/tests/test_examples.py::test_special_ioc_examples[caproto.ioc_examples.areadetector_image-DetectorGroup-class_kwargs1]
1 caproto/tests/test_examples.py::test_ioc_examples[curio-caproto.ioc_examples.rpc_function-MyPVGroup-class_kwargs6]
```

danielballan:
We finally got a couple green checkmarks on Travis.

The headline numbers from Windows stand at

```
9 failed, 1510 passed, 11 skipped, 2 xfailed, 1 xpassed, 1 warnings in 631.98 seconds 
```

danielballan:
Note to selves: that last commit has been rearranged in history to sit on time. It _may_ be the commit that fixed the deadlock in test_pyepics_pv, but if the build before it passes well, then we'll gather it's unnecessary, remove it, and then merge.

danielballan:
Reading through the 9b183f1 build, I see one build hang (timed out by Travis) on `test_pyepics_pv` and a couple other failed on it and/or the access rights test after it. Reading through 
c953169, both of those pass every time. We should keep c953169.

PR #216, opened 2018-05-25
Fix v12 support

tacaswell:
This needs some testing, but it works against a real v12 server that was failing before.

The fix is to:

 - carry the version information on the circuit (can you have mixed version channels on the same circuit?)
 - aggressively update the protocol version if we find out that it is a lower version than we asked for
 - break the API on `_BaseChanel` to get the version from the circuit

Maybe we should prevent a create channel command from going out until we have a VersionResponse?

### DIFF
--- a/caproto/_circuit.py
+++ b/caproto/_circuit.py
@@ -408,7 +408,10 @@ class _BaseChannel:
         if data_type is None:
             data_type = self.native_data_type
         if data_count is None:
-            data_count = 0
+            if self.protocol_version >= 13:
+                data_count = 0
+            else:
+                data_count = self.native_data_count
         return data_type, data_count
 
     def state_changed(self, role, old_state, new_state, command):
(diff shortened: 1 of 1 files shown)


danielballan:
I have to refresh my memory of when the VersionResponses come back. I am in favor of locking this down as much as possible, but IIRC some servers reserve the right not to bother telling you their Version more than once.

danielballan:
> Maybe we should prevent a create channel command from going out until we have a VersionResponse?

We do: https://github.com/NSLS-II/caproto/blob/master/caproto/_state.py#L46

caproto/_circuit.py
```diff
@@ -322,6 +324,16 @@ def _process_command(self, role, command):
                 raise err("priority {} does not match previously set priority "
                           "of {} for this circuit".format(command.priority,
                                                           self.priority))
+            protocol_version = min(self.protocol_version, command.version)
+            self.protocol_version = protocol_version
+            for cid, chan in self.channels.items():
```
danielballan: I think we should not bother with this loop (and the one like it below). If the same circuit gives us different VersionResponses, we should either leave existing channels alone or blow up.

danielballan: I'm arguing that it doesn't make sense for the same circuit to have more than one version. It's an open question whether any existing servers (that we need to interact with) violate this reasonable assumption anyway....

PR #226, opened 2018-05-29
FIX: Shutdown socket properly.

danielballan:
``socket.shutdown`` has a required argument. Because this happens in a cleanup
step it was not being caught but it generated error messages in the test logs.


danielballan:
Approved over private channels. Merging.

danielballan:
I should have waited for Travis. This passes locally but repeatably [hangs](https://travis-ci.org/NSLS-II/caproto/jobs/385324741#L2602) on Travis, after a suspicious series of alternative passes and failures. I do not yet understand the problem.

PR #242, opened 2018-06-03
FIX: Fix interface and port configuration.

### LINKED ISSUE #222: Respect bind address passed to servers.
As noticed by @tacaswell, we currently ignore user input from Python and just use the configuration from the environment variables. The input from Python should take precedence (or at least not be silently ignored).

danielballan:
Closes #222. ~~More accurately, it _will_ once it's ready to merge -- right now this is just docs to help me sort out what to do.~~

### DIFF
--- a/caproto/server/common.py
+++ b/caproto/server/common.py
@@ -392,7 +392,7 @@ class Context:
 
                 if known_pv:
                     # responding with an IP of `None` tells client to get IP
-                    # address from packet
+                    # address from the datagram.
                     search_replies.append(
                         ca.SearchResponse(self.port, None, command.cid, 13)
                     )
--- a/caproto/server/common.py
+++ b/caproto/server/common.py
@@ -479,6 +479,32 @@ class Context:
                 if hasattr(instance, 'server_startup') and
                 instance.server_startup is not None}
 
+    def _bind_tcp_sockets_with_consistent_port_number(self, make_socket):
+        # Find a random port number that is free on all self.interfaces,
+        # and get a bound TCP socket with that port number on each
+        # interface. The argument `make_socket` is expected to be a
+        # synchronous callable with the signature
+        # `make_socket(interface, port)` that does whatever
+        # library-specific incantation is necessary to return a bound
+        # socket or raise an IOError.
+        tcp_sockets = {}  # maps interface to bound socket
+        stashed_ex = None
+        for port in ca.random_ports(100):
+            try:
+                for interface in self.interfaces:
+                    s = make_socket(interface, port)
+                    tcp_sockets[interface] = s
+            except IOError as ex:
+                stashed_ex = ex
+                for s in tcp_sockets.values():
+                    s.close()
+                tcp_sockets.clear()
+            else:
+                break
+        else:
+            raise RuntimeError('No available ports and/or bind failed') from stashed_ex
+        return port, tcp_sockets
+
     async def tcp_handler(self, client, addr):
         '''Handler for each new TCP client to the server'''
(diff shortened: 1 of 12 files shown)


danielballan:
Do not merge. I'd like to take this opportunity to document what each of the environment variables does, which components pay attention to them and which Python parameters potentially override them.

caproto/server/common.py
```diff
@@ -446,15 +449,15 @@ def __getitem__(self, pvname):
         addresses = get_beacon_address_list()
 
         while True:
-            if self.udp_sock is None:
+            if self.beacon_sock is None:
                 await self.async_layer.library.sleep(.1)
                 continue
             beacon = ca.RsrvIsUpResponse(13, self.port, self.beacon_count,
-                                         self.host)
+                                         self.interfaces[0])  # TODO which one?
```
klauer: Important TODO here - I think it should probably indicate the IP for each interface in the beacon message.

caproto/server/server.py
```diff
@@ -968,7 +968,19 @@ def ioc_arg_parser(*, desc, default_prefix, argv=None, macros=None):
...
+    default_intf = get_server_address_list()
+    if default_intf == ['0.0.0.0']:
+        default_msg = '0.0.0.0'
+    else:
+        default_msg = (f"{' '.join(default_intf)} as specified by environment "
+                       f"variable EPICS_CAS_INTF_ADDR_LIST")
+    parser.add_argument('--interfaces', default=default_intf,
+                        nargs='+',
+                        help=(f"Interfaces to listen on. Default is "
+                              f"{default_msg}.  Multiple entires can be "
```
klauer: `entires`

klauer: Also, very nice having this in the help string.

caproto/tests/conftest.py
```diff
@@ -456,25 +451,8 @@ def server(request):
...
-                        await ctx.run()
-                    except OSError as ex:
-                        # if we get an OS error, likely due to
-                        # port already is use, try again
-                        continue
-                    else:
-                        break
-                else:
-                    raise ex
+                ctx = caproto.curio.server.Context(pvdb)
```
klauer: :+1:

caproto/trio/server.py
```diff
@@ -111,27 +116,47 @@ def __init__(self, host, port, pvdb):
...
-                for addr, port in ca.get_server_address_list(self.port):
-                    self.log.debug('Listening on %s:%d', addr, port)
+                # TODO Should this interface be configurable?
+                await self.beacon_sock.bind(('0.0.0.0', 0))
+                # Find a random port number that is free on all interfaces,
+                # and get a bound TCP socket with that port number on each
+                # interface.
+                tcp_sockets = {}  # maps interface to bound socket
+                stashed_ex = None
+                for port in ca.random_ports(100):
```
klauer: Is it possible to push this up in server common?

danielballan: I started that way and pulled it apart when I realize that trio treats binding as asynchronous whereas curio and asyncio treat it as blocking. But, reflecting on your comment, I realized there are other places where only two of the three libs use code from common (and one is special-cased). So I moved the code in the base class for curio and asyncio and special-cased trio.

caproto/server/common.py
danielballan: Does that mean we want a separate beacon loop (and `beacon_sock`) for each interface?

danielballan:
I originally suggested waiting to merge this until thorough documentation could be included. I'd rather punt that to a separate issue.

This looks good to go but I'd feel better if @klauer could look over 4a5bda0.

PR #249, opened 2018-06-04
ENH: Warn on duplicate search results in trio and curio clients.

danielballan:
Verifying warnings in pytest is easy, but verifying that WARNING log
messages were sent is harder (seems to require third-party plugins, and
the support does not look great).

So I tested this interactively. Will circle back for unit testing of
warnings in general.

### DIFF
--- a/caproto/curio/client.py
+++ b/caproto/curio/client.py
@@ -287,13 +287,23 @@ class SharedBroadcaster:
                                        command.version)
                         continue
                 if isinstance(command, ca.SearchResponse):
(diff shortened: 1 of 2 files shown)


danielballan:
I forgot to mention that the sync client _can't_ warn about duplicate search results because it immediately stops listening for responses as soon as it finds the first match. In summary:

- threading -- already has this feature
- curio and trio -- added in this PR
- sync -- can't support this feature

danielballan:
5/10 builds pass; failures are random.

PR #252, opened 2018-06-05
MNT: update header generation

### LINKED ISSUE #250: Be smarter about 'has_overflowing_field'
As in: https://github.com/epics-base/epics-base/blob/3.14/src/cas/generic/outBuf.cc#L104

danielballan:
This cleans up some unnecessarily awkwardness/verboseness in the auto-generated header code.

Closes #250 

----

*I'm taking the liberty of making a PR out of some work contributed by @klauer late last night so that Travis can spin on it while I do other stuff...*


danielballan:
I'm thinking, for example, that you could cache VersionResponse / VersionRequests and keep reusing the same instance. This is what you are alluding to?

klauer:
I meant more like `EchoRequestHeader = MessageHeader(23, 0, 0, 0, 0, 0)`.
Leaving that for a later refactor, if desirable.

PR #253, opened 2018-06-05
REF: Use QueueHandler.

### LINKED ISSUE #240: Reopen question of StreamHandler deadlocks
[This commit](https://github.com/NSLS-II/caproto/pull/215/commits/c953169a42b3b100756b67b3f06beddaf4ce8f57) fixed an intermittent deadlock in the logging StreamHandler's I/O lock by simply removing logging from the pyepics-compat layer,  relying on the logging from the threading client layer. A better approach is needed because users should be able to safely add logger to higher layers without fear of deadlocks.

danielballan:
Closes #240 (maybe) but has some downsides:

- Starts two threads at import
- Log messages related to, say, fetching a reading sometimes get processed
  after that reading has already been returned, so in interactive use (in IPython)
  some log messages get dumped after the return value, which is less nice to look at.

I find myself hoping that this is a better solution, but this seems to work.

### DIFF
--- a/caproto/_log.py
+++ b/caproto/_log.py
@@ -1,4 +1,6 @@
 import logging
+import logging.handlers
+import queue
 import sys
 try:
     import colorama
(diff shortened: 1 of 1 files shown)


danielballan:
Hmm, surprisingly this change seems to make the threading client dead lock _more_. The first two builds deadlocked -- haven't checked the others yet -- and that's more than the total number of threading client deadlocks I have ever seen before.

klauer:
Closing as this is no longer needed. I also didn't really like the out-of-order log messages vs stdout/stderr. 

Feel free to re-open if you feel otherwise.

danielballan:
Yep, never liked this. Ordered log messages are important.

PR #255, opened 2018-06-05
API: Client Contexts instantiate a fresh SharedBroadcaster by default.

### LINKED ISSUE #251: Make client Contexts’ broadcaster argument optional
...and simplify docs accordingly.

Per discussion, there might be a case where Contexts want to share the same Broadcaster, so we plan to leave them as separate classes for now, but the common case is that if you want a fresh Context you also want a fresh Broadcaster.

danielballan:
Closes #251.


danielballan:
8/10 builds passed.

PR #257, opened 2018-06-06
API: Rework sync client and CLIs

### LINKED ISSUE #139: Implement a package-standard way of spawning a repeater
* I have tried a double-fork (see `caproto._utils.spawn_daemon`) but forking does weird things from inside IPython.
* I have tried `Popen`, which works fine by assumes `caproto-repeater` is on the `PATH`. It should be, but calling the function directly smells better to me.

Both of these approaches _always_ spawn a repeater and let that repeater die if it can't bind to the repeater port. This seems wasteful, but I can't think of a better way. The only way to check whether the port is open is to bind to it, and binding to it causes it to become temporarily unavailable (until reclaimed by the OS). So this process seems problematic:

1. Bind to see if it's available.
2. It's available! Start a repeater, which will attempt to bind.

If I understand correctly, (2) can fail depending on race conditions. Better to do: bind and start a repeater or fail to bind and die. But this is admittedly wasteful and I am open to better suggestions! It could be that my understanding of how binding works is not quite right.

### LINKED ISSUE #254: Refactor sync and command-line functions.
[Updated to check off progress in #257]

- [x] Move each command-line function into a separate module with a `__main__` block to support usages with `python -m caproto.get` in the even that the entry-point aliases (`caproto-get` etc.) aren't working due to `PATH`-related issues.
- [x] In the sync client (NOT the CLIs) rename get -> read, put -> write, and monitor -> subscribe.
- [ ] Make the signatures of the sync client consistent with their counterparts in the threading client.
- [x] Have subscribe return a Subscription object (again like the threading client) that implements `block()` to run and `clear()` as a hook to stop it, from another thread, without raising `KeyboardInterrupt`.

danielballan:
Closes #254 
Closes #139


danielballan:
Things to look at with fresh eyes:

- The "poison pill" logic that supports `sub.interrupt()` isn't quite right.
- `sub.clear()` needs a fresh look
- The signatures of `read`, `write` don't quite match their threading counterparts yet.
- The `write` function should just write -- not sandwich the write between two reads. That should only happen at the CLI level, to mimic `caput`.

caproto/threading/client.py
```diff
@@ -1494,9 +1494,9 @@ def __init__(self, pv):
         self._callback_lock = threading.RLock()
 
     def add_callback(self, func):
-        # TODO thread safety
-        cb_id = self._callback_id
-        self._callback_id += 1
+        with self._callback_lock:
```
klauer: :+1: threadsafety - perhaps combine into one block, moving out all things that don't need locking?

caproto/get.py
```diff
@@ -0,0 +1,136 @@
...
+    if raw_data_type is None:
+        data_type = None
+    else:
+        assert isinstance(raw_data_type, str)
+        # ChannelType is an IntEnum.
+        # If d is int, use ChannelType(d). If string, getattr(ChannelType, d).
+        try:
+            data_type_int = int(raw_data_type)
+        except ValueError:
+            data_type = getattr(ChannelType, raw_data_type.upper())
```
klauer: Did you know you can use `ChannelType[raw_data_type.upper()]` as well?

caproto/monitor.py
```diff
@@ -0,0 +1,127 @@
...
+    parser.add_argument('--no-repeater', action='store_true',
+                        help=("Do not spawn a Channel Access repeater daemon "
+                              "process."))
+    parser.add_argument('--priority', '-p', type=int, default=0,
+                        help="Channel Access Virtual Circuit priority. "
+                             "Lowest is 0; highest is 99.")
+    parser.add_argument('--timeout', '-w', type=float, default=1,
+                        help=("Timeout ('wait') in seconds for server "
+                              "responses."))
+    exit_group.add_argument('--duration', type=int, default=None,
```
klauer: Nice! Maybe float though?

caproto/monitor.py
```diff
@@ -0,0 +1,127 @@
...
+        for pv_name in args.pv_names:
+            sub = subscribe(pv_name,
+                            mask=mask,
+                            priority=args.priority)
+            cb = functools.partial(callback, pv_name)
+            sub.add_callback(cb)
+            cbs.append(cb)  # Hold ref to keep cb from being garbage collected.
+            subs.append(sub)
+        # Wait to be interrupted by KeyboardInterrupt.
+        block(*subs, duration=args.duration, timeout=args.timeout,
```
klauer: Rather cool we can do this, even with the sync client.

caproto/sync/client.py
```diff
@@ -492,22 +309,31 @@ def monitor(*pv_names, callback, mask=None, timeout=1,
...
                 selector.register(sock, selectors.EVENT_READ)
-            logger.debug('Continuing until SIGINT is received....')
+            if duration is None:
+                logger.debug('Continuing until SIGINT is received....')
             while True:
                 events = selector.select(timeout=0.1)
+                if deadline is not None and time.time() > deadline:
+                    logger.debug('Deadline reached.')
+                    return
+                global poison_pill
```
klauer: :skull:

caproto/sync/repeater.py
```diff
@@ -199,8 +201,29 @@ def run(host='0.0.0.0'):
...
+                     'one.')
+        return
+    # We will now spawn a repeater in a subprocess at the same address as sock.
+    # Make the address reusable so that, if the OS does not clean up sock
+    # before the subprocess tries to bind to it, there is no conflict.
+    try:
+        reuse = socket.SO_REUSEADDR
+    except AttributeError:
+        warnings.warn("SO_REUSEADDR is not supported on this platform.")
+    sock.setsockopt(socket.SOL_SOCKET, reuse, 1)
```
klauer: Should this be in the `else` block?

caproto/sync/client.py
tacaswell: :ghost:

danielballan:
> I'm not super stoked about cluttering the top level directory even more, can they be stuffed down one more level (either into a cli package or back into sync)?

@klauer suggested putting them at the top level so that the usage

```
python -m caproto.get ...
```

is brief. On the other hand, we have `caproto-get`, and the `python -m` variant was just intended to be a backup in the event that the entry points aren't working due to `PATH` / virtual environment weirdness, so maybe it's not that important that it be brief. I'm 50/50 on this.

caproto/monitor.py
danielballan: Bah. Copy-pasted from `--maximum`. Fixed.

caproto/get.py
danielballan: Neat! I did not.

caproto/threading/client.py
danielballan: There's no way around this ordering (nothing to move outside the lock) but I guess it's probably better to acquire once and do a little extra work in the locked state that to thrash the lock twice in quick succession.

caproto/_commands.py
```diff
@@ -110,17 +111,23 @@ def has_metadata(data_type):
...
     if has_metadata(data_type):
         md_payload = DBR_TYPES[data_type].from_buffer(buffer)
         md_size = ctypes.sizeof(DBR_TYPES[data_type])
     else:
         md_payload = b''
         md_size = 0
-    # Use payload_size to strip off any right-padding that may have been added
-    # to make the byte-size of the payload a multiple of 8.
-    data_payload = memoryview(buffer)[md_size:md_size + payload_size]
+    # Use data_count to strip off any right-padding that may have been added
```
klauer: This diff looked scarier than it actually is. 👍

PR #264, opened 2018-06-09
FIX: a couple of small but important fixes

tacaswell:


caproto/threading/client.py
```diff
@@ -594,12 +602,18 @@ def _retry_unanswered_searches(self):
...
+                    yield ca.SearchRequest(it[0], search_id, 13)
+                    it[-1] = t
+
             with self._search_lock:
-                items = list(self.unanswered_searches.items())
-            requests = (ca.SearchRequest(name, search_id, 13)
-                        for search_id, (name, _) in items)
+                items = list((search_id, it)
+                             for search_id, it in self.unanswered_searches.items()
+                             if (t - it[-1]) > RETRY_SEARCHES_PERIOD / 2)
```
danielballan: Just curious -- what's the `/ 2` for? Is this is an expedient way to try to make sure that we retry _at least_ that often, even accounting for the time it takes to go through this loop?

caproto/threading/client.py
```diff
@@ -550,19 +552,25 @@ def command_loop(self):
                     cid = command.cid
                     try:
                         with self._search_lock:
-                            name, queue = unanswered_searches.pop(cid)
+                            name, queue, _ = unanswered_searches.pop(cid)
                     except KeyError:
                         # This is a redundant response, which the EPICS
                         # spec tells us to ignore. (The first responder
                         # to a given request wins.)
-                        if name in self.search_results:
```
danielballan: Yeppp. This would be a `NameError` on `name`, a problem I saw in logs yesterday. Gotta fix this (and much else) in the async clients too.

caproto/threading/client.py
tacaswell: Yes.

caproto/threading/client.py
tacaswell: Worse, if it has been through once it _works_ but tells you the wrong PV name!

PR #265, opened 2018-06-09
REF/STY/DOC: Follow-up on #264

danielballan:
I merged #264  a little too quickly -- there were ``flake8`` errors. This fixes
those and does other minor stuff.


danielballan:
One build passed, three otheres failed on one random failure (not all the same random failure). Killing the build and merging to make space in the pipeline for the next one....

PR #271, opened 2018-06-11
DOC: Many updates in prep for v0.1.0

danielballan:


danielballan:
Docs build without warnings! Merging in the spirit of "something is better than nothing" (for docs). Please review published docs.

PR #278, opened 2018-06-12
MNT: switch examples to using asyncio by default

### LINKED ISSUE #225: Consider moving built-in IOC examples to asyncio
I like trio better, but it would be cool to have the ioc examples runnable with zero deps.

tacaswell:
Pin several examples to curio

closes #225

### DIFF
--- a/caproto/asyncio/server.py
+++ b/caproto/asyncio/server.py
@@ -12,10 +12,44 @@ class ServerExit(Exception):
     ...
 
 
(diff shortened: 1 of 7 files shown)


danielballan:
I would still prefer if this PR came with an implementation of `ThreadsafeQueue`.

klauer:
Agreed @danielballan - we should be able to put one of those together without much difficulty.

PR #279, opened 2018-06-12
ENH: Add Batch context.

danielballan:


caproto/threading/client.py
```diff
@@ -1789,6 +1792,110 @@ def __del__(self):
             pass
 
 
+class Batch:
```
klauer: This `Batch` syntax/functionality is even nicer than I had imagined - nice work!

caproto/threading/client.py
```diff
@@ -1789,6 +1792,110 @@ def __del__(self):
...
+    """
+    Accumulate requests and then issue them all in batch.
+
+    Parameters
+    ----------
+    timeout : number or None
+        Number of seconds to wait before ignoring late responses. Default
+        is 2.
```
klauer: Maybe in this case a linear function, proportional to number of requests, would be reasonable?

caproto/threading/client.py
```diff
@@ -1789,6 +1792,110 @@ def __del__(self):
...
+
+    Examples
+    --------
+    Read some PVs in batch and stash the readings in a dictionary as they
+    come in.
+
+    >>> readings = {}
+
+    >>> results = {}
+    >>> stash_result = lambda name, response: results.update({name: response.data})
```
klauer: flake8: no lambda! 😄 
```
def stash_result(name, response):
    results[name] = response.data
```

^ seems a bit more clear for a basic example, I think.

caproto/threading/client.py
```diff
@@ -1789,6 +1792,110 @@ def __del__(self):
...
+    timeout : number or None
+        Number of seconds to wait before ignoring late responses. Default
+        is 2.
+
+    Examples
+    --------
+    Read some PVs in batch and stash the readings in a dictionary as they
+    come in.
+
+    >>> readings = {}
```
klauer: `readings` unused

caproto/threading/client.py
```diff
@@ -1789,6 +1792,110 @@ def __del__(self):
...
+
+    >>> results = {}
+    >>> stash_result = lambda name, response: results.update({name: response.data})
+
+    >>> with Batch() as b:
+    ...     for pv in pvs:
+    ...         b.read(pv, functools.partial(stash_result, pv.name))
```
klauer: I think it's good as-is, but thinking of alternatives: what if `read` also gave you a `Future` of sorts for getting the result of each request? A dictionary? Hmm...

caproto/threading/client.py
danielballan: Good suggestion of syntax!

caproto/threading/client.py
danielballan: I think the asyncio variety should return a Future. Trio doesn't _have_ futures, so not sure about that. I think we should add _something_ once we see how this shakes out in other frameworks.

caproto/threading/client.py
danielballan: Better do this on all the methods then. Punting to https://github.com/NSLS-II/caproto/issues/284 (but plan to address today)

danielballan:
Tests passed; failed on flake8 again. Pushed fix to test code, reran tests locally to verify I didn't break the tests I touched. Merging.

PR #280, opened 2018-06-12
API: Make ioc_arg_parser default_prefix default to ''.

danielballan:
As I walked Ke through writing a caproto IOC, I wondered whether we can make
``default_prefix`` optional, just to remove number of things you have to know
to get an IOC up.

I don't feel strongly and defer to others' sense of whether it is a good
idea to let people write an IOC without forcing them to invent a default prefix.
Just a thought.

### DIFF
--- a/caproto/server/server.py
+++ b/caproto/server/server.py
@@ -935,7 +935,7 @@ class PVGroup(metaclass=PVGroupMeta):
         return value
 
 
-def ioc_arg_parser(*, desc, default_prefix, argv=None, macros=None):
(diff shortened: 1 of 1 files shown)


klauer:
I could be convinced if you think it adds to ease of initial setup. That said, knowing the prefix exists and gets pre-pended may be a good starting introduction to how the `PVGroup` structure works.

danielballan:
I think we should probably _not_ merge this unless some thinks it's definitely a good idea, and I am not that person.

PR #282, opened 2018-06-12
REF: try using coloredlogs module

klauer:
Just seeing what this looks like, in place of vendoring the logging from tornado.

### DIFF
--- a/caproto/_log.py
+++ b/caproto/_log.py
@@ -1,190 +1,31 @@
 import logging
-import sys
+import warnings
+
 try:
-    import colorama
-    colorama.init()
+    import coloredlogs
 except ImportError:
-    colorama = None
-try:
-    import curses
-except ImportError:
-    curses = None
-
-
-__all__ = ('color_logs', 'plain_log_handler', 'color_log_handler')
-
-# The LogFormatter is adapted light from tornado, which is licensed under
-# Apache 2.0. See other_licenses/ in the repository directory.
-
-# A bunch of py2-safe stuff from Tornado that we do not need in py3 land.
-basestring_type = unicode_type = _safe_unicode = str
-
-
-def _stderr_supports_color():
-    try:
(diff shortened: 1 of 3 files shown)


danielballan:
Not sure we want to merge this, but big :+1: for at least trying it out so we know what we like.

teddyrendahl:
Curious as to why this was a no go

danielballan:
It's a fair question. From the beginning we've gone a little out of our way to make all of caproto's core features work with zero external dependencies beyond Python itself. We took the trouble to make numpy optional, falling back on Python's built-in `array` module until numpy is present. We have an `asyncio`-based implementation of the server and, of course, a `threading` implementation of the client --- in addition to other implementations that we find more interesting but require external libraries to use.

This lightness is nice to have in contexts like installing caproto on RaspberryPi, and I guess I just find it entertaining/satisfying as an experiment. If we ever find ourselves needing something big that Python doesn't provide, I would be in favor of importing it rather than reinventing it, but so far Python's built-ins have had all the essential pieces. I vendor-ed (and @klauer simplified) Tornado's color log implementation, and since it's not _that_ large or complex, I don't think it's worth breaking our zero-deps streak in order to use coloredlogs. I would feel differently if caproto already had a bunch of deps.

danielballan:
I appreciate that the above position is a little eccentric. :- D

teddyrendahl:
Didn't mean for it to come off as a pointed question 😄 

We have been happy with `coloredlogs` so I was just curious if there was something about the library itself that turned you away. I appreciate the desire for a pure Python standard lib implementation regardless of the alleged eccentricity 👍

danielballan:
Yeah, no worries, didn't take it personally at all -- I just figured this was a good time to record our thinking in general so I can refer back to it later if/when it comes up again with another module.

teddyrendahl:
😄

PR #287, opened 2018-06-12
ENH: add full support for the sync filter

### LINKED ISSUE #267: Implement the 'sync' ChannelFilter in the server.
We parse it, as of #263, but the server ignores it.

### LINKED ISSUE #300: Number of connected clients logged by server jumps up by 2 sometimes.
...which leaves the total count wrong, growing over time inaccurately.

klauer:
States are defined on the `PVGroup` level, inherited from the parent if not specified. Currently set in the class definition, but this probably should be redone - looking for suggestions.

~Note: No short-hand syntax support just yet~
Adds simple `PvProperty` for boolean values with ['No', 'Yes'] as the enum strings.

Closes #267 
Closes #300

### DIFF
--- a/caproto/_data.py
+++ b/caproto/_data.py
@@ -503,7 +503,6 @@ class ChannelData:
 
     def post_state_change(self, state, new_value):
         snapshots = self._snapshots[state]
-        snapshots.clear()
         if new_value:
             # We have changed from false to true.
             snapshots['while'] = self
@@ -660,6 +659,12 @@ class ChannelData:
                                                  self.string_encoding)
             await self.write_metadata(publish=False, **metadata_dict)
 
+        if self._fill_at_next_write:
+            snapshot = copy.deepcopy(self)
+            for state, mode in self._fill_at_next_write:
+                self._snapshots[state][mode] = snapshot
+            self._fill_at_next_write.clear()
+
         # Send a new event to subscribers.
         await self.publish(flags, (old, new))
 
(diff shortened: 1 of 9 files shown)


danielballan:
Rebased.

caproto/_data.py
```diff
@@ -436,15 +443,77 @@ def __init__(self, *, alarm=None,
...
+        self._fill_at_next_write = list()
+
+    def __setstate__(self, val):
+        self.timetstamp = val['timestamp']
+        self._alarm = val['alarm']
+        self.string_encoding = val['string_encoding']
+        self.reported_record_type = val['reported_record_type']
+        self._data = val['data']
+        self.string_encoding = val['string_encoding']
+        self._data = val['data']
```
klauer: Duplicate line. Consider also perhaps: ```self.__dict__.update(...)``` for all of the `val` items.

caproto/_data.py
```diff
@@ -621,33 +699,39 @@ def _is_eligible(self, ss, flags, pair):
...
+                for data_type, sub_specs in data_types.items():
+                    eligible = tuple(ss for ss in sub_specs
+                                     if self._is_eligible(ss))
+                    if not eligible:
+                        continue
+                    if sync is None:
+                        channel_data = self
+                    else:
+                        try:
+                            channel_data = self._snapshots[sync.s][sync.m]
```
klauer: Should we consider renaming the `s` and `m` to have some meaning, despite the filter-JSON names?

PR #290, opened 2018-06-12
FIX: Enforce limits.

danielballan:
A bunch of tests were running PV (served by caproto) past their limits
and had to be updated.


danielballan:
Tests passed; pushed an extra commit to make flake8 happy.

PR #291, opened 2018-06-12
ENH: Channel Access timestamp filter

teddyrendahl:
First caproto PR 🎉 

This should be fairly simple, but I am pretty unfamiliar with this stuff so please correct me.

## Questions
1. `metadata` is a reference so I do not want to mutate that information so it looked like I was forced to create a new `metadata` object.
2. Am I guaranteed to get a `DBR_TIME_LONG` or do I need to introspect the class and create a new one that way?
```
(caproto) [trendahl@psbuild-rhel7-02  caproto]$ caget -a simple:A
simple:A                       2018-06-12 13:17:21.927088 1  
(caproto) [trendahl@psbuild-rhel7-02  caproto]$ caget -a 'simple:A.{"ts": {}}'
simple:A.{"ts": {}}            2018-06-12 13:33:48.677312 1  
```

### DIFF
--- a/caproto/server/common.py
+++ b/caproto/server/common.py
@@ -1,9 +1,8 @@
-from collections import defaultdict, deque, namedtuple
+from collections import defaultdict, deque, namedtuple, ChainMap
 import logging
 import time
 import caproto as ca
 from caproto import apply_arr_filter, get_environment_variables
-from caproto._dbr import DBR_TIME_LONG
 
 
 class DisconnectedCircuit(Exception):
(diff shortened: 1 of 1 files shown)


caproto/server/common.py
```diff
@@ -211,7 +212,14 @@ def get_db_entry():
                 command.data_type, user_address=self.circuit.address)
             # This is a pass-through if arr is None.
             data = apply_arr_filter(chan.channel_filter.arr, data)
-
+            # If the timestamp feature is active swap the timestamp.
+            # Information must copied because not all clients will have the
+            # timestamp filter
+            if chan.channel_filter.ts:
+                now = ca.TimeStamp.from_unix_timestamp(time.time())
+                metadata = DBR_TIME_LONG(status=metadata.status,
```
klauer: I think we'd have to ensure that `command.data_type` is in `time_types` (https://github.com/NSLS-II/caproto/blob/master/caproto/_dbr.py#L1023), otherwise the request doesn't make sense. Then use `DBR_TYPES[type]` to get the appropriate `DBR_TYPE_*` structure.

teddyrendahl:
Thanks. Generalized for different data types now.

teddyrendahl:
Where and how to add a test? I think there is some pretty simple logic but not sure how to include it.

```python
def tests_timestamp_filter():
      proc_time = read('my:pv').metadata.timestamp
      res = read('my:pv.{"ts": {}}')
      assert res.metadata.timestamp != proc_time
```

danielballan:
Merged! Welcome to the conspiracy!

I had a stale tab open and your comment didn't load until after I merged. Let's add a `test_channel_filters.py`. I'll get it started....

PR #295, opened 2018-06-13
BLD: Display a helpful error message if setup.py detects an old Python version.

danielballan:


danielballan:
Tested with OSX system pip:

```
$ pip install .
Processing /Users/dallan/Documents/Repos/caproto
    Complete output from command python setup.py egg_info:

    Caproto does not support Python 2.x, 3.0, 3.1, 3.2, 3.3, 3.4, or 3.5.
    Python 3.6 and above is required. Check your Python version like so:

    python --version

    This may be due to an out of date pip. Make sure you have pip >= 9.0.1.

    pip install --upgrade pip


    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in /private/var/folders/c5/qrjg0b6j0bd4cq8g7knhjn2h0000gn/T/pip-req-build-yJhdZi/
```

danielballan:
Good pass rate. Merging as noncontroversial.

PR #303, opened 2018-06-14
ENH: Accept scalar or str/bytes in write.

danielballan:
The sync client already does this. (Checked interactively.)

This PR makes the threading client do the same and adds a test.

### DIFF
--- a/caproto/threading/client.py
+++ b/caproto/threading/client.py
@@ -9,6 +9,7 @@
 # - restart subscriptions
 # - ThreadPoolExecutor for processing user callbacks on read, write, subscribe
 import array
+from collections import Iterable
 import concurrent.futures
 import errno
 import functools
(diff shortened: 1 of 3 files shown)


caproto/threading/client.py
```diff
@@ -1460,6 +1461,10 @@ def write(self, data, *, wait=True, callback=None, timeout=2,
             Requested number of values. Default is the channel's native data
             count.
         """
+        if not isinstance(data, Iterable) or isinstance(data, (str, bytes)):
+            data = [data]
+        if data and isinstance(data[0], str):
+            data = [val.encode('latin-1') for val in data]
```
klauer: I think we have a string encoding environment variable somewhere...

caproto/threading/client.py
```diff
@@ -1873,6 +1878,10 @@ def write(self, pv, data, callback=None, data_type=None, data_count=None):
...
```
klauer: Ditto

PR #305, opened 2018-06-14
ENH: Add more records for mock_record option in high-level server

klauer:
Adds all from recordwhat excluding: `mca`, `scaler`

There are a good number of menus (enums) that I have yet to convert, so they are shown as `NotImplementedMenu` at the moment. Can add that to the TODO list at a low priority.

Auto-conversion now also links known parent attributes with `_link_parent_attribute`.

Generated with simply:
```python
from caproto.server.conversion import record_to_high_level_group_code
import recordwhat

record_types = recordwhat.get_known_records()

for record_type in sorted(record_types - {'mca', 'scaler'}):
    print()
    print()
    for line in record_to_high_level_group_code(record_type, skip_fields=['VAL']):
        print(line)
```


danielballan:
Caproto welcomes recordwhat and is excited about the EPICS "road trip" they will go on together.

The first block of Travis builds were green, save for one common random failure, so I canceled the rest. Merging!

PR #307, opened 2018-06-14
FIX: tweak barcode example and fix doc string

tacaswell:


danielballan:
This suggests #308. I'll merge as is for v0.1.0.

Issue #310, opened 2018-06-15
Monitor benchmark suggestion


danielballan:
From Michael Davidsaver (recalled from conversation by me, I hope accurately):

> Serve one PV that updates at a slow, steady rate. Serve a second PV whose update rate ramps slowly. Observe at what point the client has to start dropping messages from the ramping PV and, further, verify that it is processing the update from the steady PV at an even rate. This checks that the client is "fair."

hhslepicka:
That is right! I can confirm this talk and the issue reflects the discussed problem. ;)

PR #312, opened 2018-06-15
ENH: add a 'chirp' server

tacaswell:


danielballan:
Related to #310

klauer:
Very cool. I'd like to add this to the benchmarks, but could probably do so after merging. Is it ready?

tacaswell:
I had thoughts of making this a 'triggered' chirp instead of continuous but never got to it.

PR #317, opened 2018-06-16
FIX: Issue beacons faster at startup.

### LINKED ISSUE #311: Issue Beacons more quickly at startup.
The CA spec is clear about how to do this, but our servers don't implement it yet.

danielballan:
Closes #311


danielballan:
Lotta random failures on this but they do not seem correlated.

PR #319, opened 2018-06-19
FIX: Enforce quotas per subscription under high load.

danielballan:
Tested interactively, it correctly starts dropping updates for a specific Subscription on the floor if a given client is behind.

If there is a backlog of EventAddResponses to send, it starts batching them into blocks to send at once. This is tested interactively too and seems to work well.

### DIFF
--- a/caproto/curio/server.py
+++ b/caproto/curio/server.py
@@ -41,12 +41,15 @@ class VirtualCircuit(_VirtualCircuit):
 
     def __init__(self, circuit, client, context):
         super().__init__(circuit, client, context)
-        self.command_queue = curio.Queue()
+        self.command_queue = curio.Queue(ca.MAX_COMMAND_BACKLOG)
         self.new_command_condition = curio.Condition()
         self.pending_tasks = curio.TaskGroup()
+        self.events_on = curio.Event()
+        self.subscription_queue = curio.Queue(ca.MAX_TOTAL_SUBSCRIPTION_BACKLOG)
 
     async def run(self):
         await self.pending_tasks.spawn(self.command_queue_loop())
+        await self.pending_tasks.spawn(self.subscription_queue_loop())
 
     async def _on_disconnect(self):
         """Executed when disconnection detected"""
@@ -60,6 +63,12 @@ class VirtualCircuit(_VirtualCircuit):
     async def _wake_new_command(self):
         async with self.new_command_condition:
             await self.new_command_condition.notify_all()
+
+    async def get_from_sub_queue_with_timeout(self, timeout):
+        # Timeouts work very differently between our server implementations,
+        # so we do this little stub in its own method.
+        # Returns weakref(EventAddResponse) or None
+        return await curio.ignore_after(timeout, self.subscription_queue.get)
 
 
 class Context(_Context):
--- a/caproto/curio/server.py
+++ b/caproto/curio/server.py
@@ -29,6 +29,18 @@ class UniversalQueue(curio.UniversalQueue):
         return await super().get()
 
 
+class QueueFull(Exception):
+    ...
+
+
+# Curio queues just block if they are full. We want one that raises.
+class QueueWithFullError(curio.Queue):
+    async def put(self, item):
+        if self.full():
+            raise QueueFull
+        await super().put(item)
+
+
 class CurioAsyncLayer(AsyncLibraryLayer):
     name = 'curio'
     ThreadsafeQueue = UniversalQueue
@@ -41,11 +53,13 @@ class VirtualCircuit(_VirtualCircuit):
 
     def __init__(self, circuit, client, context):
         super().__init__(circuit, client, context)
-        self.command_queue = curio.Queue(ca.MAX_COMMAND_BACKLOG)
+        self.QueueFull = QueueFull
+        self.command_queue = QueueWithFullError(ca.MAX_COMMAND_BACKLOG)
         self.new_command_condition = curio.Condition()
         self.pending_tasks = curio.TaskGroup()
         self.events_on = curio.Event()
-        self.subscription_queue = curio.Queue(ca.MAX_TOTAL_SUBSCRIPTION_BACKLOG)
+        self.subscription_queue = QueueWithFullError(
+            ca.MAX_TOTAL_SUBSCRIPTION_BACKLOG)
 
     async def run(self):
         await self.pending_tasks.spawn(self.command_queue_loop())
--- a/caproto/server/common.py
+++ b/caproto/server/common.py
@@ -301,9 +301,10 @@ class VirtualCircuit:
                     self.log.warning("High EventAddResponse load. Dropped "
                                      "%d responses.", num_expired)
                 if len_commands > 1:
-                    self.log.info("High EventAddResponse load. Batch size: "
-                                  "%d commands (%d bytes).",
-                                  len_commands, commands_bytes)
+                    self.log.info(
+                        "High EventAddResponse load. Batching with %f-second "
+                        "latency. Batch size: %d commands (%d bytes).",
+                        HIGH_LOAD_TIMEOUT, len_commands, commands_bytes)
                 await self.send(*commands)
             except DisconnectedCircuit:
                 await self._on_disconnect()
--- a/caproto/server/common.py
+++ b/caproto/server/common.py
@@ -70,6 +70,7 @@ class VirtualCircuit:
         self.unexpired_updates = defaultdict(
             lambda: BoundedSet(maxlen=ca.MAX_SUBSCRIPTION_BACKLOG))
         # Subclasses are expected to define:
+        # self.QueueFull = ...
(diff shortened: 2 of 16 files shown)


danielballan:
Before I started hardening this with tests, does the implementation look reasonable?

tacaswell:
How strong of a guarantee is it that `set.pop` is ordered?

klauer:
Interesting! Will get back to you with actual useful input.

Following up on @tacaswell's doubt, [`set.pop`](https://docs.python.org/3.6/library/stdtypes.html?highlight=set#frozenset.pop) definitely does not follow the insertion order: `set.pop? -> Remove and return an arbitrary set element.`

danielballan:
I have found that it is ordered in Python 3.6 but not ordered in previous versions. I suspect that the dict refactor in Python also made sets ordered, but I am still looking for documentation of this and any formal policy about it.

I think that this change to caproto might be fine in either case because randomly dropping updates is also a reasonable policy, and all of this is within the domain of "implementation details" according to our conversation with Jeff Hill. But of course we should understand and document the behavior accurately before merging.

danielballan:
OK, I have something conclusive. Sets are still not ordered in general:

```py
>>> s = set(['a', 'b', 'c'])
>>> s
{'c', 'a', 'b'}
>>> s.pop()
'c'
```

but in the case of integers (at least integers at the scale we care about, which is 2**16 and below) they are ordered because of this:

```py
>>> hash(1)
1
>>> hash(2)
2
>>> hash(65536)
65536
```

When the subscription ID counter loops around from 65535 to 0, this will cause the "new" news to be popped off before the "old" news, until all of the old news (6553X) is either successfully sent or dropped on the floor.

I think I can live with that, provided that we document it clearly in code comments. The alternative is to accept considerably more overhead in bookkeeping. Reactions?

klauer:
I suppose the `set` "ordering" is sufficient, given the usage. Did you consider (instead of `get_from_sub_queue_with_timeout`) passively monitoring the delta time between updates? To someone who didn't implement anything (me), it seems like that would lead to a simpler solution.

danielballan:
> When the subscription ID counter loops around from 65535 to 0, this will cause the "new" news to be popped off before the "old" news, until all of the old news (6553X) is either successfully sent or dropped on the floor.

> I think I can live with that, provided that we document it clearly in code comments. The alternative is to accept considerably more overhead in bookkeeping. Reactions?

I have had a change of heart about this. The "considerably more overhead in bookkeeping isn't so bad, and I think it could be harmful to suddenly, silently switch from systematically tossing old news (ideal) to systematically tossing new news. We've established that either is OK --- it's an implementation detail --- but _switching_ modes systematically seems bad. 8f14228 fixes by implementing an `OrderedBoundedSet`.

> Did you consider (instead of `get_from_sub_queue_with_timeout`) passively monitoring the delta time between updates? To someone who didn't implement anything (me), it seems like that would lead to a simpler solution.

Hmm. That does seem potentially simpler. Is there an obvious place to get this time _from_? Not all subscriptions must include a time field. Do you have in mind capturing the time of insertion into the queue and putting that into the queue item alongside the update?

danielballan:
> Did you consider (instead of `get_from_sub_queue_with_timeout`) passively monitoring the delta time between updates? To someone who didn't implement anything (me), it seems like that would lead to a simpler solution.

Actually, now that I load this load back into my brain, I don't think that simpler approach is viable. The problem to be solved here is: I have a batch of responses ready to send. No one knows if the device is about to reproduce yet another response _any millisecond now_. (Not even the device can tell us yet -- it depends what happens in the physical world!) I want to wait for a short interval to find out if more updates are coming. If one arrives, I should add it to the batch and keep waiting until my batch gets large. If not, I should time out and send the batch I have.

tacaswell:
Who invited code-quality back to the party....

caproto/curio/server.py
```diff
@@ -29,6 +29,18 @@ def get(self):
...
 
 
+class QueueFull(Exception):
+    ...
+
+
+# Curio queues just block if they are full. We want one that raises.
+class QueueWithFullError(curio.Queue):
+    async def put(self, item):
+        if self.full():
```
tacaswell: race condition, but not clear there is a way around that here....

danielballan: I think it's not because there are no threads and no awaits.

caproto/server/common.py
```diff
@@ -24,6 +37,41 @@ class DisconnectedCircuit(Exception):
...
+    def add(self, item):
+        self._data[item] = None
+        if len(self) > self.maxlen:
+            self.pop()
+
+    def update(self, items):
+        self._data.update(dict.fromkeys(items))
+        overage = len(self) - self.maxlen
+        if overage > 0:
+            for _ in range(overage):
```
tacaswell: see a micro-optomization here as 

```python
for obs in islice(iter(self._data), overage)):
    self._data.pop(obs)
```

caproto/server/common.py
```diff
@@ -567,10 +741,28 @@ def __getitem__(self, pvname):
...
+                # client "new news" instead of "old news".
+
+                # This is an OrderedBoundedSet, a set with a maxlen, containing
+                # only commands for this particular subscription.
+                circuit.unexpired_updates[sub.subscriptionid].add(command)
+
+                # This is a queue with the commands from _all_ subscriptions on
+                # this circuit.
+                try:
+                    await circuit.subscription_queue.put(weakref.ref(command))
```
tacaswell: should this queue be responsible for evicting the 'old news'?

caproto/curio/server.py
tacaswell: fair point!

danielballan:
One random failure on one build. Noting it here in case we want to circle back for it later:

```
______________ test_multiple_subscriptions_one_server[epics-base] ______________
ioc = namespace(name='Waveform and standard record IOC', prefix='c1ab7f77:', process=<subprocess.Popen object at 0x7fbc7e809...m', 'str': 'c1ab7f77:str', 'int': 'c1ab7f77:int', 'int2': 'c1ab7f77:int2', 'int3': 'c1ab7f77:int3'}, type='epics-base')
context = <Context searches_pending=7 circuits=0 pvs=7 idle=0>
    def test_multiple_subscriptions_one_server(ioc, context):
        pvs = context.get_pvs(*ioc.pvs.values())
        for pv in pvs:
>           pv.wait_for_connection(timeout=10)
caproto/tests/test_threading_client.py:275: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
self = <PV name='c1ab7f77:waveform' priority=0 (searching....)>, args = ()
kwargs = {'timeout': 10}
pv = <PV name='c1ab7f77:waveform' priority=0 (searching....)>, raw_timeout = 10
timeout = 9.999987379999993, deadline = 710.707443253, ready = False, i = 0
    @functools.wraps(func)
    def inner(self, *args, **kwargs):
        if isinstance(self, PV):
            pv = self
        elif isinstance(self, Subscription):
            pv = self.pv
        else:
            raise TypeError("ensure_connected is intended to decorate methods "
                            "of PV and Subscription.")
        # This `2` matches the default in read, write, wait_for_connection.
        raw_timeout = timeout = kwargs.get('timeout', 2)
        if timeout is not None:
            deadline = time.monotonic() + timeout
        with pv._in_use:
            pv._usages += 1
            # If needed, reconnect. Do this inside the lock so that we don't
            # try to do this twice. (No other threads that need this lock
            # can proceed until the connection is ready anyway!)
            if pv._idle:
                # The Context should have been maintaining a working circuit
                # for us while this was idle. We just need to re-create the
                # Channel.
                ready = pv.circuit_ready.wait(timeout=timeout)
                if not ready:
                    raise TimeoutError(f"Could not connect within "
                                       f"{raw_timeout}-second timeout.")
                with pv.component_lock:
                    cm = pv.circuit_manager
                    cid = cm.circuit.new_channel_id()
                    chan = ca.ClientChannel(pv.name, cm.circuit, cid=cid)
                    cm.channels[cid] = chan
                    cm.pvs[cid] = pv
                    pv.circuit_manager.send(chan.create())
        try:
            for i in range(CIRCUIT_DEATH_ATTEMPTS):
                # On each iteration, subtract the time we already spent on any
                # previous attempts.
                if timeout is not None:
                    timeout = deadline - time.monotonic()
                ready = pv.channel_ready.wait(timeout=timeout)
                if not ready:
>                   raise TimeoutError(f"Could not connect within "
                                       f"{raw_timeout}-second timeout.")
E                                      TimeoutError: Could not connect within 10-second timeout.
caproto/threading/client.py:91: TimeoutError
```

danielballan:
Now that `[caproto]` version of that same test failed on the same build....

danielballan:
And finally on the third try, the board is green.

PR #321, opened 2018-06-20
CI: see if pypy3.5 supports enough of the 3.6 parts we use

tacaswell:


tacaswell:
`enum.auto()` is apparently new in 3.6, might keep poking at this locally...

PR #329, opened 2018-07-02
DOC: Correct inaccurate docstring.

danielballan:
This prompts the question, should we change the API of the access_rights_callback
to be the same --- to take the PV instance as the first argument? I think that
we should.


klauer:
👍 I think so.

PR #330, opened 2018-07-23
FIX: Bind broadcast socket to random port in standard range.

danielballan:


danielballan:
The right fix here might be to actually remove the hard-coded range from `random_ports` and refactor the code that calls that, not to _add_ the hard-coded range to more places as I originally thought when I opened this PR.

Various OS-es have different ephemeral port ranges, and caproto should defer to the OS if at all possible. https://www.cymru.com/jtk/misc/ephemeralports.html h/t @klauer

klauer:
I'm 50/50 on this as mentioned before (would like to rely on OS-assigned ports). @tacaswell tie breaker?

danielballan:
@klauer I'm on board with fixing this the right way. Will leave PR open as a reminder.

tacaswell:
Where did we used to bind this socket?

klauer:
The first `sendto` causes it to be bound automatically to a random port, so there were no explicit `bind` calls (if I recall correctly).

danielballan:
> if I recall correctly

Yes, that is right.

klauer:
I believe this has been superseded by #543 which we believe to be "correct" wrt epics-base.

PR #335, opened 2018-08-07
FIX: Update internal calls to new sig of _accces_rights_changed.

danielballan:
Bug was introduced in 7f7f2e43, #329


danielballan:
Fixes failures locally, self-mering as a hotfix.

PR #337, opened 2018-09-07
ENH: In CLIs, accept -vvvvvvvv(etc) as a synonym for -vvv.

danielballan:
I remember someone (@tacaswell) remarking that it is annoying that ``-vvvv`` is
less verbose than ``-vvv``, and it's easy enought to make ``-v`` a "count"
thing rather than a "store_true" thing. This is also exactly how ansible's
verbosity CLI flag is handled, as a point of precedent.


tacaswell:
I am very excited about this change!

PR #339, opened 2018-09-10
FIX: Deadband filter now works without numpy.

danielballan:
It was comparing network-endian arrays rather than host-endian.


tacaswell:
why did it _ever_ work?!

danielballan:
It didn't ever work. We threw the channel filter work in at the last minute (for the Chicago talk) and in our haste we wrongly interpreted the failures as flaky, intermittent ones, not noticing that they exclusively and repeatedly affect the "no numpy" builds.

PR #340, opened 2018-09-15
CI/test suite tweaks

klauer:


tacaswell:
you need 

```
dist: xenial
sudo: true
```

for the 3.7 enviroment

Issue #328, opened 2018-06-28
Investigate ophyd test failures


danielballan:
It looks like we introduced a regression at some point, only just now noticed in [this ophyd build](https://travis-ci.org/NSLS-II/ophyd/jobs/397632649).

The proximate error is:

```py
dtype = <ChannelType.CHAR: 4>, values = [b'/data/2018/06/28/\x00']
    def python_to_epics(dtype, values, *, byteswap=True, convert_from=None):
        'Convert python builtin values to epics CA'
        # NOTE: ignoring byteswap, storing everything as big-endian
        if dtype == ChannelType.STRING:
            return DbrStringArray(values).tobytes()
    
>       return np.asarray(values).astype(type_map[dtype])
E       ValueError: invalid literal for int() with base 10: '/data/2018/06/28/'
```

Detailed excerpt from build logs:

<details>

```
_________ test_fstiff_plugin[caproto-None-/data/%Y/%m/%d-None-False1] __________
root = None, wpath = '/data/%Y/%m/%d', rpath = None, check_files = False
    @pytest.mark.parametrize('root,wpath,rpath,check_files',
                             ((None, '/data/%Y/%m/%d', None, False),
                              (None, '/data/%Y/%m/%d', None, False),
                              ('/data', '%Y/%m/%d', None, False),
                              ('/data', '/data/%Y/%m/%d', '%Y/%m/%d', False),
                              ('/', '/data/%Y/%m/%d', None, False),
                              ('/tmp/data', '/data/%Y/%m/%d', '%Y/%m/%d', True)
                              ))
    def test_fstiff_plugin(root, wpath, rpath, check_files):
        fs = DummyFS()
        fs2 = DummyFS()
        if check_files:
            fh = pytest.importorskip('databroker.assets.handlers')
    
        class FS_tiff(TIFFPlugin, FileStoreTIFF,
                      FileStoreIterativeWrite):
            pass
    
        class MyDetector(SingleTrigger, SimDetector):
            tiff1 = Cpt(FS_tiff, 'TIFF1:',
                        write_path_template=wpath,
                        read_path_template=rpath,
                        root=root, reg=fs)
        target_root = root or '/'
        det = MyDetector(prefix, name='det')
        det.read_attrs = ['tiff1']
        det.tiff1.read_attrs = []
    
>       det.stage()
ophyd/tests/test_areadetector.py:338: 
[...]

PR #341, opened 2018-09-24
Fix support for CHAR dtype.

danielballan:
This is a fix for #328. It raises the possibility that there may be other related failures, missed by our test suite. I plan to take this opportunity to heavily revise our datatype-conversion code, but that is a lengthy project, so I submit this hot fix for now. This fixes the ophyd test failures.


danielballan:
...But this introduces new caproto failures. Don't bother reviewing it. More to come.

danielballan:
Superceded by commits tagged onto #351

PR #343, opened 2018-10-04
Add simulated temperature controller

tacaswell:


danielballan:
This edits the auto-generated code. Maybe it would be better to edit `generate_headers.py`.

caproto/ioc_examples/thermo_sim.py
```diff
@@ -0,0 +1,76 @@
...
+    @setpoint.putter
+    async def setpoint(self, instance, value):
+        self._T0 = time.monotonic()
+        return value
+
+
+if __name__ == '__main__':
+    parser, split_args = template_arg_parser(
+        default_prefix='thermo:',
+        desc='One PV updates at a steady rate; another "chirps" (accelerating updates).')
```
danielballan: Copy/paste relic?

tacaswell: indeed.

tacaswell:
did not know that file was auto-generated, will just drop that commit.

caproto/ioc_examples/thermo_sim.py
```diff
@@ -0,0 +1,85 @@
...
+        self.period = period
+
+    readback = pvproperty(value=[0], dtype=float, read_only=True)
+
+    setpoint = pvproperty(value=[100], dtype=float)
+    K = pvproperty(value=[10], dtype=float)
+    omega = pvproperty(value=[(np.pi)], dtype=float)
+    Tvar = pvproperty(value=[10], dtype=float)
+
+    @readback.startup
```
klauer: Can you use the newer `.scan` feature here instead?

caproto/ioc_examples/thermo_sim.py
```diff
@@ -0,0 +1,85 @@
...
+    @setpoint.putter
+    async def setpoint(self, instance, value):
+        self._T0 = time.monotonic()
+        return value
+
+
+if __name__ == '__main__':
+    parser, split_args = template_arg_parser(
+        default_prefix='thermo:',
+        desc=dedent(Thermo.__doc__))
```
klauer: Agreed with @danielballan - though I get the desire to change docs in one place and one place only, this seems a bit excessive...

tacaswell:
@klauer what do you want tested about this?  Just that it starts?

I have a copy-paste fail, I get complaints, I prevent any future copy-paste I get complaints ;)

The main benefit I see to putting all of `__doc__` into the help string is that it documents the relationship between the PVs that are going to be exposed.  It is much nicer to be able to ask `--help` for that information than to have to go find the source.  It does not currently contain a `Parameters` section. 

I don't see how to use `.scan` and make the scan time run-time configurable.

klauer:
Yeah, good point about the scan rate being tied to a PV. I'm going to propose a fix for that later, it's again on my TODO list.

For testing, it just belongs in this list, nothing more:
https://github.com/NSLS-II/caproto/blob/master/caproto/tests/test_examples.py#L338

awalter-bnl:
I have pushed some changes after discussion with @tacaswell . Essentially I have changed `setpoint` and `readback` to `SP` and `I` respectively so that they match the PV structure of the Eurotherm controller this was written or testing with.

caproto/ioc_examples/thermo_sim.py
```diff
@@ -33,32 +33,32 @@ def __init__(self, *args, period, **kwargs):
         self._T0 = time.monotonic()
         self.period = period
 
-    readback = pvproperty(value=[0], dtype=float, read_only=True)
+    I = pvproperty(value=[0], dtype=float, read_only=True)
```
klauer: You could have also specified `name="SP"` and left it as `setpoint` for the property name ([ref](https://github.com/NSLS-II/caproto/blob/d1610fa259cb7ff83b44c5c7eed1044faa48eef1/caproto/server/server.py#L288))

danielballan: :+1: for that approach, to keep the names readable by humans.

awalter-bnl: change has been made !-)

danielballan:
@awalter-bnl Do you still have your bash history from this work? There are a bunch of extra commits in here now, so something might be funny about your git workflow.

awalter-bnl:
@danielballan I have had this issue regularly, but it is not every time so I don't really understand where I am going wrong. 
I will take a look at this first thing tomorrow, and see. 

FYI The steps I took prior to this PR were:
`$git checkout -b enh_thermo_sim`
`$git pull tacaswell enh_thermo_sim` (this created the first commit which is the merge commit)
Make the first set of changes
`$git status`
`$git add caproto/ioc_examples/thermo_sim.py`
`$git commit -m 'some string' --author 'Andrew Walter <awalter@bnl.gov'`
`$git push tacaswell enh_thermo_sim`
then enter git username and password
wait for comments
Make the second lot of changes
`$git status`
`$git add caproto/ioc_examples/thermo_sim.py`
`$git commit -m 'some string' --author 'Andrew Walter <awalter@bnl.gov'`
`$git push tacaswell enh_thermo_sim`
then enter git username and password

danielballan:
>  this created the first commit which is the merge commit

This should not have required a merge. What was the state of git before you pulled?

awalter-bnl:
I am not sure what the state was prior to the pull.

I used the pull command above, it automatically opened a vi editor and asked for a 'merge commit message' I wrote something then closed and saved the editor and it did the rest, is that where I am going wrong?

tacaswell:
I'm going to stand by docstring here

```
17:34 $ python thermo_sim.py --help
usage: thermo_sim.py [-h] [--prefix PREFIX] [-q | -v] [--list-pvs]
                     [--async-lib {asyncio,curio,trio}]
                     [--interfaces INTERFACES [INTERFACES ...]]
                     [--period PERIOD]

Simulates (poorly) an oscillating temperature controller.

Follows :math:`T_{output} = T_{var} exp^{-(t - t_0)/K} sin(ω t) + T_{setpoint}`

The default prefix is `thermo:`

Readonly PVs
------------

I -> the current value

Control PVs
-----------

SP -> where to go
K -> the decay constant
omega -> the oscillation frequency
Tvar -> the scale of the oscillations

optional arguments:
  -h, --help            show this help message and exit
  --prefix PREFIX
  -q, --quiet           Suppress INFO log messages. (Still show WARNING or
                        higher.)
  -v, --verbose         Verbose mode. (Use -vvv for more.)
  --list-pvs            At startup, log the list of PV names served.
  --async-lib {asyncio,curio,trio}
                        Which asynchronous library to use. Default is curio.
  --interfaces INTERFACES [INTERFACES ...]
                        Interfaces to listen on. Default is 0.0.0.0. Multiple
                        entries can be given; separate entries by spaces.
  --period PERIOD       The update rate of the readback

```

Could be formatted better, but I like that it gives notes about what the PVs mean.

50/50 on waiting moving to use `SCAN` before Ken's PR goes in.

tacaswell:
re-based on current master to pick up the threading test fixes.

danielballan:
> Could be formatted better, but I like that it gives notes about what the PVs mean.

OK, I'm convinced.

danielballan:
> 50/50 on waiting moving to use SCAN before Ken's PR goes in.

Now that #350 is in, are you up for refactoring this to use it?

caproto/tests/test_examples.py
```diff
@@ -343,6 +343,7 @@ def find_put_value(pv):
     [('caproto.ioc_examples.all_in_one', 'MyPVGroup',
       dict(macros={'macro': 'expanded'})),
      ('caproto.ioc_examples.chirp', 'Chirp', {'ramprate': 0.75}),
+     ('caproto.ioc_examples.thermo_sim', 'Thermo', {'period': 0.75}),
```
klauer: `period` kwarg here is causing test failures

tacaswell: you beat me to fixing this by about 4 minutes

danielballan:
@tacaswell, thanks for this `__doc__` cleverness and for convincing @klauer and me that it was a good idea.

@awalter-bnl, congrats on your first commit to caproto.

PR #344, opened 2018-10-07
ENH: allow customization of fields on a per-record basis

klauer:
Although this PR adds some more confusing code to the magical high-level server, I think its benefits could be worth it. Take a look at the examples to see how it's relatively easy to hook into the pvproperties of the mocked fields. 

There are two options:
1. Specify `mock_record`, then use the decorator: `@record.fields.<field_name_or_attribute_name>.putter`
2. Subclass and modify the `RecordFieldGroup` (either making a new record type or overriding the internal one) - now possible thanks to the rewritten record registry


caproto/ioc_examples/mocking_records.py
```diff
@@ -14,11 +16,35 @@ class RecordMockingIOC(PVGroup):
             # an exception in the putter:
             raise ValueError('Invalid value!')
 
+    # It's also possible to modify some of the behavior of fields on a per-
+    # record basis. The following function is called whenever B.RVAL is put to:
```
tacaswell: should be `A.RVAL`?

tacaswell:
how does the functionality overlap with FLNK?  I think we need both (and epics-base provides both)?

PR #345, opened 2018-10-07
Fix epics_ca_addr_list warning

klauer:
Still on the path to better CI...


danielballan:
Restarted one job after this (random, I think) failure:

```
FAIL caproto/tests/test_threading_client.py::test_multithreaded_many_read[epics-base-1-iter2]
```

PR #346, opened 2018-10-08
TST: increase barrier wait time due to travis slowness

klauer:
Based on what I've seen locally with barrier breakage using 100 to 300+ threads, it seems that the time needed to complete these tests on travis is significantly greater than when run locally. Let's see if this makes multithreaded tests pass more repeatably.

### DIFF
--- a/caproto/threading/client.py
+++ b/caproto/threading/client.py
@@ -1906,16 +1906,11 @@ class Subscription(CallbackHandler):
     def _subscribe(self, timeout=2):
         """This is called automatically after the first callback is added.
         """
-        with self.pv.component_lock:
-            with self._callback_lock:
-                has_callbacks = bool(self.callbacks)
-            if has_callbacks:
-                cm = self.pv.circuit_manager
-                ctx = cm.context
-                with ctx.subscriptions_lock:
-                    ctx.subscriptions_to_activate[cm].add(self)
-                ctx.activate_subscriptions_now.set()
-            return has_callbacks
+        cm = self.pv.circuit_manager
+        ctx = cm.context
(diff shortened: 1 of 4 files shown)


danielballan:
One build passed. Each of the other 3 failed one or more variant of 
`caproto/tests/test_threading_client.py::test_multithreaded_many_subscribe`

caproto/threading/client.py
```diff
@@ -1369,10 +1359,10 @@ def _process_command(self, command):
             pv.connection_state_changed('connected', chan)
             # If we have just revived an existing PV whose
             # VirtualCircuit died and reconnected, we are now ready to
-            # reinstate its Subsciprtions. If this is a new PV, it
-            # won't have any Subscriptions.
+            # reinstate its Subsciprtions.
```
klauer:
```suggestion
            # reinstate its Subscriptions.
```
... what a typo!

caproto/threading/client.py
```diff
@@ -1849,14 +1845,14 @@ def removed(_):
...
+        with self.callback_lock:
             cb_id = self._callback_id
             self._callback_id += 1
             self.callbacks[cb_id] = ref
         return cb_id
 
     def remove_callback(self, token):
-        with self._callback_lock:
+        with self.callback_lock:
             self.callbacks.pop(token, None)
```
klauer: `pop` is atomic

tacaswell: But you do not want this to fire while something else is iterating over the callbacks etc.

danielballan: Right.

PR #356, opened 2018-10-14
CI: pip needs -U to obey pins in install_requires.

danielballan:


danielballan:
Trio [correctly specifies](https://github.com/python-trio/trio/blob/master/setup.py#L77) that it needs a recent version of attrs, but pip is ignoring it.

danielballan:
The reason, then, that this was not reproducible locally was that Travis seeds the test environment with a version of pytest and, thereby, attrs. In a local environment, all of this is installed from scratch, and the dependencies are resolved differently.

danielballan:
Got an annoying number of random failures (only 2/4 jobs passed on the first try) but none of the failures were repeatable or suggested any connection to the trio upgrade.

klauer:
Great find!

PR #358, opened 2018-10-14
DOC: add iOS/Pythonista installation notes

klauer:
EPICS on your iPhone has never been so easy.


tacaswell:
This made me laugh quite a bit.

Issue #357, opened 2018-10-14
Refactor out use of deprecated trio.Queue


danielballan:
See trio release notes: https://trio.readthedocs.io/en/latest/history.html#trio-0-9-0-2018-10-12

danielballan:
As part of this work, once it's ready, we should unpin trio: https://github.com/NSLS-II/caproto/pull/355/commits/7b711f066f23e5cc6459fe56cca244383e5c5b1a

PR #362, opened 2018-10-15
REF: replace (deprecated) trio queues with trio memory channels

klauer:
This PR looks more drastic than it is. It effectively breaks up the "common" server loops into 2 methods - splitting queue access and parse/respond:
1. `*_loop()` Reference loop implementation, functioning just like current master's loops - grabbing from a queue and calling `*_iteration()` (2)
2. `*_iteration()` "Single iteration" - the portion which takes a single command and does something with it

I think I prefer this compared to the current implementation regardless of the trio memory channel refactor.

Addresses #357

### DIFF
--- a/caproto/asyncio/server.py
+++ b/caproto/asyncio/server.py
@@ -92,7 +92,7 @@ class VirtualCircuit(_VirtualCircuit):
         self._sq_task = None
         self._write_tasks = ()
 
-    async def get_from_sub_queue_with_timeout(self, timeout):
+    async def get_from_sub_queue(self, timeout=None):
         # Timeouts work very differently between our server implementations,
         # so we do this little stub in its own method.
         fut = asyncio.ensure_future(self.subscription_queue.get())
--- a/caproto/curio/server.py
+++ b/caproto/curio/server.py
@@ -91,10 +91,13 @@ class VirtualCircuit(_VirtualCircuit):
         async with self.new_command_condition:
             await self.new_command_condition.notify_all()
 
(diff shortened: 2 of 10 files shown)


klauer:
I couldn't effectively rebase due to all of the changes that happened in the server between the time this was submitted and now, so it was just redone.

@danielballan any further thoughts on how the subscription queues could be handled (within the bounds of trio channels)?

klauer:
I think this is probably ready. It refactors away all client and server `trio.Queue`s.

danielballan:
Thanks, @klauer. Apologies that I am just now noticing your question from December 12. Will review carefully in the new year.

danielballan:
We seem to have caught an internal trio error:

```
trio.TrioInternalError: internal error in trio - please file a bug!
```

Context:

https://travis-ci.org/NSLS-II/caproto/jobs/469798681#L3363-L3599

klauer:
Is the cause of that internal error just pytest-timeout canceling the hung test?

danielballan:
Sure looks that way. Will throw some cycles at this today.

danielballan:
> Will throw some cycles at this today.

...or hopefully next week. Busy with deployments!

danielballan:
9774b5a got 4 green builds on the first try and 8074b87 is now far enough through Travis that I'm convinced I didn't do anything dumb there (just a simple change to `test-requirements.txt`). Merging -- thanks @klauer! I think this is worth a new micro release, which I will do shortly.

PR #366, opened 2018-10-22
FIX: Address corner cases and high-load tuning with acctst

### LINKED ISSUE #323: Pass acctst
A tracking issue for any problems we encounter while running the caproto servers against the regression tester in epics-base, `acctst`.

danielballan:
See #323 for progress updates.

### DIFF
--- a/caproto/server/common.py
+++ b/caproto/server/common.py
@@ -298,7 +298,7 @@ class VirtualCircuit:
                         now - deadline + latency_limit)
 
                 # Ensure at the last possible moment that we don't send
-                # respones for Subscriptions that have been canceled at some
+                # responses for Subscriptions that have been canceled at some
                 # time after the response was queued. The important thing is
                 # that no EventAddResponse be sent after the corresponding
                 # EventCancelResponse.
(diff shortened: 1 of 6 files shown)


caproto/server/common.py
```diff
@@ -314,7 +316,18 @@ def __init__(self, circuit, client, context):
                         "High load. Batched %d commands (%dB) with %.4fs latency.",
                         len_commands, commands_bytes,
                         now - deadline + HIGH_LOAD_TIMEOUT)
-                await self.send(*commands)
+
+                # Ensure at the last possible moment that we don't send
+                # respones for Subscriptions that have been canceled at some
```
klauer: `responses`

caproto/server/common.py
```diff
@@ -464,9 +478,8 @@ def get_db_entry():
                         status=write_status,
                         data_count=db_entry.length
                     )
-
-                if client_waiting:
-                    await self.send(response_command)
+                    if client_waiting:
```
tacaswell: might as well move this up to include creating the response?  Do we have to turn the crank an creating that even if the client is not waiting for it back?

tacaswell: This is the first time in python that I have missed the extra scoping you can get with `{}` in c++...

caproto/server/common.py
```diff
@@ -704,29 +704,21 @@ def __getitem__(self, pvname):
...
-                # A specific sub_id has been specified, which means this update
-                # is due to a *new* Subscription, and should only go to that
-                # specific Subscription. Other Subscriptions have already been
-                # sent this update and should not receive it again. Find which
-                # Subscription that is....
+                # A specific Subscription has been specified, which means this
+                # specific update was prompted by Subscription being new, not
+                # prompted by a new value. The update should only be sent to
+                # that specific Subscription.
+                subs = (sub,)
```
tacaswell: super nit-pick, make this a list to match the type used above?  Or replace the above with a tuple comprehension?

danielballan: Reasonable.

PR #369, opened 2018-10-25
FIX: use by default minimum MTU likely to be seen in the wild

klauer:
TODO: consider determining MTU per interface dynamically in the future.
There may not be portable solutions for this.

Linux: https://gist.github.com/nzjrs/8934855

(Seen on ophyd areadetector testing on OSX - part of an upcoming PR)

Edit: 
(1) a bit of pork in 232a169, pyepics compat in threading client was calling the incorrect connection state changed method, meaning that `epics.PV` connection callbacks were never run.

(2) new flake8 is much more strict

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,4 +1,4 @@
 [flake8]
-ignore = E402
+ignore = E402,W504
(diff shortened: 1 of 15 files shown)


danielballan:
3/4 passed. One failure that was a timeout.

PR #370, opened 2018-10-26
ENH: More debug prints in threading client.

danielballan:
The current DEBUG logs do not include PV names except at channel
creation time. This adds DEBUG logs to channel-level logging that
include the PV name, read request/response, write request/response, and
event response. It omits event request because that is managed
asynchrously by the circuit and is typically of lower operational
interest.

image

### DIFF
--- a/caproto/threading/client.py
+++ b/caproto/threading/client.py
@@ -1455,7 +1455,7 @@ class PV:
         deadline = time.monotonic() + timeout if timeout is not None else None
         ioid_info['deadline'] = deadline
         self.circuit_manager.send(command)
-        self.log.debug(f"{self.name!r}: {command}")
+        self.log.debug("%r: %r", self.name, command)
         if not wait:
             return
 
@@ -1475,7 +1475,7 @@ class PV:
                 f"{self.name!r} within {timeout}-second timeout."
             )
 
-        self.log.debug(f"{self.name!r}: {ioid_info['response']}")
+        self.log.debug("%r: %r", self.name, ioid_info['response'])
         return ioid_info['response']
 
     @ensure_connected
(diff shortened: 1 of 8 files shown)


caproto/threading/client.py
```diff
@@ -1455,6 +1455,7 @@ def read(self, *, wait=True, callback=None, timeout=2, data_type=None,
         deadline = time.monotonic() + timeout if timeout is not None else None
         ioid_info['deadline'] = deadline
         self.circuit_manager.send(command)
+        self.log.debug(f"{self.name!r}: {command}")
```
klauer: While I really like f-strings, for logging there's no better choice than the old-style string formatting. This will cause command repr to be generated even if the log level is not set such that this message would go through.

danielballan: :man_facepalming: Knew that, literally opened a terminal to play with the details of that implementation yesterday, and then did not apply it....

caproto/threading/client.py
```diff
@@ -1775,14 +1779,15 @@ def _unsubscribe(self, timeout=2):
...
+    def process(self, command):
         # TODO here i think we can decouple PV update rates and callback
         # handling rates, if desirable, to not bog down performance.
         # As implemented below, updates are blocking further messages from
         # the CA servers from processing. (-> ThreadPool, etc.)
 
-        super().process(*args, **kwargs)
-        self.most_recent_response = (args, kwargs)
+        super().process(command)
+        self.log.debug(f"{self.pv.name!r}: {command}")
```
klauer: Missed one

danielballan:
Threadling client timeout errors now include the repr of the PV that timed out. While I'm here, I went in and changes `XError` -> `CaprotoXError` in almost all public code.

PR #372, opened 2018-10-28
CI: initial azure test

klauer:
Auto-generated azure-pipelines.yml

Separate (nonfunctional) work going on in klauer/caproto@azure

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,0 +1,124 @@
+# Python package
+# Create and test a Python package on multiple Python versions.
+# Add steps that analyze code, save the dist with the build record, publish to a PyPI-compatible index, and more:
+# https://docs.microsoft.com/azure/devops/pipelines/languages/python
+# vi: sw=2 ts=2 sts=2 expandtab
+
+variables:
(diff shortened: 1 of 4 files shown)


mrakitin:
The error is caused by the absence of the `requirements.txt` file, which is used here: https://github.com/NSLS-II/caproto/pull/372/files#diff-fec826feae04e51c0d94076385408bdcR25. Should it be `test-requirements.txt` instead?

danielballan:
I think @mrakitin is right.

klauer:
You're both right but also wrong - as the initial PR message above makes note, this is just an autogenerated azure configuration straight from the GitHub app. 

I'm working on it in a separate branch and will push here when ready to avoid the annoying spam that's normally associated with CI-related PRs.

klauer:
OK, I've given up on this for now. It's only in a somewhat working state, but not worth wasting more of my time. 

Feel free to pick up where I left off.

klauer:
Why are we still mucking around with travis when Azure looks like this?

image

Yes, the tests fail - but they do on travis, too, in much less obvious ways.

tacaswell:
leroy jenkins.

PR #375, opened 2018-10-30
ENH: Add search backoff, cancellation, and manual search_now().

### LINKED ISSUE #246: Back off retrying SearchRequests and allow user to cancel requests.
Currently, if I search for a typo-ed PV name in the threading client, the Context will keep searching for it for the life time of the context, and there is no way to tell it to stop (short of mutating its internal `unanswered_searches` dict....).

We should do both of these:

- [ ] Back off if a given SearchRequest keeps failing. Consider the situation when one server with 1000s of PVs goes offline. The vain search for those PVs will clog up the client and slow down searches for other PVs.
- [ ] Provide some public API for saying "nevermind, I don't care about this PV anymore." The `go_idle()` API might be sufficient, but I don't think it's available/applicable to a PV in the 'searching...' state in its current form.

### LINKED ISSUE #365: Fix possible bug in threading client's re-connection.
I haven't had time to chase this down, but anecdotally I have seen reconnection fail between 10-50% of the time. I think the client sends a Create message too soon, never gets a reply, and just lands there. It should go back to square one.

### LINKED ISSUE #371: Clients should watch for beacon anomalies
See Slide 8 [here](https://epics.anl.gov/docs/APS2014/05-CA-Concepts.pdf).

danielballan:
Submitted for discussion. Tested interactively.

Closes #365 
Closes #371 
Closes #246

### DIFF
--- a/caproto/_constants.py
+++ b/caproto/_constants.py
@@ -23,7 +23,7 @@ SEARCH_MAX_DATAGRAM_BYTES = 1472
 
 # How long to wait between EchoRequest and EchoResponse before concluding that
 # server is unresponsive.
-RESPONSIVENESS_TIMEOUT = 2  # seconds
+RESPONSIVENESS_TIMEOUT = 5  # seconds
 
 MAX_TOTAL_SUBSCRIPTION_BACKLOG = 10000  # total per circuit not per subscription
 MAX_SUBSCRIPTION_BACKLOG = 1000  # per subscription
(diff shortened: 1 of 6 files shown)


klauer:
Nice! An important addition.

caproto/_constants.py
```diff
@@ -21,12 +21,9 @@
 # Max of ethernet and 802.{2,3} MTU 1500 - 20(IP header) - 8(UDP header)
 SEARCH_MAX_DATAGRAM_BYTES = 1472
 
-# Servers send beacons at some maximum interval. ("Maximum delay between
-# beacons will be limited by server specified parameter, but is commonly 15
-# seconds.") Servers can be presumed dead and dropped by circuits and repeaters
-# after some interval ("usually 30 seconds"). Reference:
-# https://epics.anl.gov/docs/CAproto.html#secVCUnresponsive
-SERVER_MIA_PRESUMED_DEAD = 60  # seconds
```
danielballan: Note about this deletion: We have a standard, spec-specified value for this in `_utils.py` already, `EPICS_CA_CONN_TMO`.

PR #379, opened 2018-11-08
FIX: empty array checks should use len() explicitly

klauer:
`if not np.array([])` is either deprecated or does not work


danielballan:
I would expect this to fail because bool of a numpy array fails. It asks you to use all or any to differentiate your meaning.

PR #383, opened 2018-11-10
**DO_NOT_MERGE** thread lock debugging / ophyd stage

klauer:
Here's the results as of 950e735 - after reducing unnecessary locking and introducing `TCP_NODELAY`:
```
$ pip install git+https://github.com/klauer/ophyd@load_cl
$ ipython stage_test.py

pyepics
2.39 ms ± 45.7 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
caproto
3.92 ms ± 175 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

### DIFF
--- a/caproto/_broadcaster.py
+++ b/caproto/_broadcaster.py
@@ -95,8 +95,8 @@ class Broadcaster:
         -------
         commands : list
         """
-        self.log.debug("Received datagram from %r with %d bytes.",
-                       address, len(byteslike))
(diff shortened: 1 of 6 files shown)


klauer:
Not portable or used in epics-base, but perhaps worth looking into for Linux: https://baus.net/on-tcp_cork/

klauer:
Closing in favor of #387 - will keep this branch for future performance debugging purposes.

PR #384, opened 2018-11-12
ENH: default do float PVs in spoof_beamline

tacaswell:


danielballan:
Two builds has one (different) random failure each, unrelated to this change. Merging without re-running.

PR #386, opened 2018-11-12
ENH: implement value-based alarms

tacaswell:


caproto/_data.py
```diff
@@ -751,13 +758,69 @@ def __init__(self, *, value, units='',
...
+                    await self.alarm.write(status=hi_status,
+                                           severity=hi_severity)
+                    return False
+
+            await self.alarm.write(status=AlarmStatus.NO_ALARM,
+                                   severity=AlarmSeverity.NO_ALARM)
+            return True
+
+        # this is HIHI and LOLO limits
+        ok = await limit_checker(val,
```
klauer: Can you pull these out into not-async functions and `await` on `alarm.write` only once?

At some point, we should start worrying more about where we `await` unnecessarily, giving the event loop a chance to switch to some other task.

caproto/ioc_examples/mocking_records.py
```diff
@@ -3,26 +3,38 @@
...
+                   units="mm",
+                   precision=3)
 
     @B.putter
     async def B(self, instance, value):
         if value == 1:
-            # Mocked record will pick up the alarm status simply by use raising
-            # an exception in the putter:
+            # Mocked record will pick up the alarm status simply by
+            # use raising an exception in the putter:
```
klauer:
```suggestion
            # us raising an exception in the putter:
```

caproto/ioc_examples/mocking_records.py
```diff
@@ -3,26 +3,38 @@
...
             raise ValueError('Invalid value!')
 
-    # It's also possible to modify some of the behavior of fields on a per-
-    # record basis. The following function is called whenever A.RVAL is put to:
+    # It's also possible to modify some of the behavior of fields on a
```
klauer: At the risk of being nitpicky, can we forego submitting formatting changes when the original source is already PEP8-compliant?

```
In [2]: len("    # It's also possible to modify some of the behavior of fields on a per-")
Out[2]: 75
In [3]: len("    # It's also possible to modify some of the behavior of fields on a")
Out[3]: 70
```

caproto/server/records.py
```diff
@@ -3641,7 +3473,7 @@ class HistogramFields(RecordFieldGroup):
 
 
 @register_record
-class LonginFields(RecordFieldGroup):
+class LonginFields(RecordFieldGroup, _Limits):
```
klauer: I have not checked the other changes to ensure that switching in `_Limits` doesn't affect anything.

However, it's obvious here that it's incorrect: `longin` limits should _not_ be `ChannelType.DOUBLE`, they should be `ChannelType.LONG`.

caproto/ioc_examples/mocking_records.py
tacaswell: I was just trusting emacs paragraph re-flow

caproto/_data.py
tacaswell: Isn't more await better to give it more chances to multi-task?

caproto/server/records.py
```diff
@@ -727,11 +695,6 @@ class AcalcoutFields(RecordFieldGroup):
         doc='Number of Elements',
         read_only=True)
     _link_parent_attribute(display_precision, 'precision')
-    _link_parent_attribute(hihi_alarm_limit, 'upper_alarm_limit')
-    _link_parent_attribute(high_alarm_limit, 'upper_warning_limit')
-    _link_parent_attribute(low_alarm_limit, 'lower_warning_limit')
-    _link_parent_attribute(lolo_alarm_limit, 'lower_alarm_limit')
-    _link_parent_attribute(low_operating_range, 'lower_ctrl_limit')
```
tacaswell: @klauer This was missing the high_operating_range links, I assumed that was an error in the initial generation?

caproto/server/records.py
```diff
@@ -1572,11 +1544,6 @@ class CalcoutFields(RecordFieldGroup):
     output_specification = pvproperty(
         name='OUT', dtype=ChannelType.STRING, doc='Output Specification')
     _link_parent_attribute(display_precision, 'precision')
-    _link_parent_attribute(hihi_alarm_limit, 'upper_alarm_limit')
-    _link_parent_attribute(high_alarm_limit, 'upper_warning_limit')
-    _link_parent_attribute(low_alarm_limit, 'lower_warning_limit')
-    _link_parent_attribute(lolo_alarm_limit, 'lower_alarm_limit')
-    _link_parent_attribute(low_operating_range, 'lower_ctrl_limit')
```
tacaswell: dittor on missing the high_operating_range link

tacaswell:
ok, I did some aggressive history re-writing and now this is ready to be reviewed commit-by-commit.  Dropped touching the Longin/Longout recrods at all for now and re-organized to pretend I got the logic right on the first try :wink:

caproto/_data.py
klauer: I'd argue that we want more atomicity in simple operations such as this - there is no good reason to task switch out of this.

Thoughts @danielballan ?

caproto/server/records.py
klauer: ~fully~ 99.9% auto-generated - I did the links manually, and of course I screwed them up! Great catch.

caproto/ioc_examples/mocking_records.py
danielballan: Does it touch lines that you don't otherwise edit?

tacaswell:
The tests are failing on search failure for the fields....

caproto/_data.py
danielballan: > I'd argue that we want more atomicity in simple operations such as this - there is no good reason to task switch out of this.

Agreed. If there is a series of operations that we know to be fast (i.e. no I/O) then there is no reason to switch tasks in the middle. That just adds overhead.

Seems like we are in agreement now, in any case.

caproto/ioc_examples/mocking_records.py
klauer: Resolved (for now) after the most recent rebase. Arbitrary sub-79 char wrapping discussion will continue in other channels...

PR #389, opened 2018-11-13
FIX: data picklability

klauer:
OK - this could be squashed as the first 2 commits are using a completely different approach than the last commit. I think the new approach is much better as I noted below.

Tests are passing, ready for review.

### DIFF
--- a/caproto/_data.py
+++ b/caproto/_data.py
@@ -77,10 +77,10 @@ class ChannelAlarm:
             severity_to_acknowledge=severity_to_acknowledge,
             alarm_string=alarm_string)
 
-    def __setstate__(self, val):
+    def __setstate__(self, state):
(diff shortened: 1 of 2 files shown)


danielballan:
Wonderful! I subscribe to Raymond Hettinger's mandate that if you really must define a custom type if your library, you ought to make it pickle-able.

klauer:
So, in testing this and remembering how `__getstate__`/`__setstate__` is so unwieldy, I stumbled upon [`__getnewargs_ex__`](https://docs.python.org/3/library/pickle.html#object.__getnewargs_ex__ ). It seems vastly superior to me in this use case. 

I'm happy to switch it back if I'm missing something though!

danielballan:
Nice find with `__getnewargs_ex__`!

PR #390, opened 2018-11-15
Add an editorconfig.

danielballan:
This should tell vim/emacs/whatever to line-wrap at 79. Notice that
we only _enforce_ wrapping at 115 to allow for the occasional situation
where a longer line is easier to read. But we encourage wrapping at 79.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,5 @@
+[*.py]
+indent_style = space
+indent_size = 4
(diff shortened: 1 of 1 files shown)


danielballan:
The modern ones do. Vim and emacs and other older ones do with a plugin. Details: https://editorconfig.org/#overview

PR #391, opened 2018-11-16
CI: remove CircleCI

klauer:


mrakitin:
What's wrong with CircleCI? I wanted to add it to one of my projects, but would like to learn why not to use it. Thanks.

klauer:
It's actually pretty good, but travis is better integrated with GitHub. One major feature circle provides is the ability to SSH into the container to debug stuff - which is _soo_ very useful.

caproto CI is difficult enough that managing more than one is not worth our time or effort, sadly. We're looking to move it to azure so we can test on OSX/Windows/Linux all at the same time, but I'm not having much luck with it (and it does not appear to have SSH support).

mrakitin:
Thanks for the nice explanation, @klauer! Azure pipelines seem very attractive. I'll give it a try (for another project first).

PR #393, opened 2018-11-17
ENH: formula IOC example: dynamically create an IOC based on a CLI "formula" parameter

tacaswell:


klauer:
It's a neat example! What do you want to do with this, @tacaswell?

caproto/ioc_examples/dynamic_calc.py
```diff
@@ -0,0 +1,58 @@
...
+
+    @formula.putter
+    async def formula(self, instance, value):
+        c, names = extract_names(value)
+
+        self.c = c
+        self.names = names
+
+        self._pvdb.clear()
+        self._pvdb.update({f'{self.prefix}{n}': ChannelDouble(value=0)
```
bl0x: Why does this have to be ChannelDouble and not pvproperty? I've tried with pvproperty and got an error.

danielballan: Hi @bl0x! It's a good question. The `pvdb` is the lowest-level interface for specifying IOCs in caproto. It maps PV names (strings) to `ChannelData` (and its subclasses, such as `ChannelDouble`).

`PVGroup` and `pvproperty` belong to a higher-level interface that, at the end of the day, also produce a `pvdb` of `ChannelData` objects. But it's generally a nicer interface to use for complex applications.

What you have tried to do is _mix_ the two, and that's not possible. If you are hand-altering the `pvdb` dict, directly you're obliged to use `ChannelData`. If you are using `pvproperty`, you must rely on `PVGroup` to build your `pvdb` for you.

tacaswell:
Although, I think the case could be made that we should push the processing of the output on every put to the input PVs....

caproto/ioc_examples/formula_ioc.py
```diff
@@ -0,0 +1,122 @@
...
+        super().__init__(*args, **kwargs)
+        self.code_object = code_object
+        self.var_names = var_names
+
+    @proc.putter
+    async def proc(self, instance, value):
+        ret = await self.compute(instance)
+        await self.output.write(ret)
+
+        return 0
```
danielballan: Should this catch any errors from compute and return 1?

tacaswell:
Right, but that is hard because we need to trigger the processing _after_ the put is complete but we do not currently have a good hook for that.

caproto/ioc_examples/formula_ioc.py
tacaswell: I don't think we have any sematics around return non-zero to mean failure on putters (as the return value of this is what is actually written to the channel).  If this raises, it will be caught by https://github.com/NSLS-II/caproto/blob/0f96128eb764054070b31e4a8c0e9b0e54407873/caproto/_data.py#L473-L486
and sets an alarm on proc.

We probably should make all of the PVs share the same alarm...

tacaswell: They now all share an alarm (thanks @klauer for `alarm_group`!).

danielballan:
Are you happy with the state of this, @tacaswell?

tacaswell:
Yes, it exposes a hook we don't have, but as a PoC it is ok.

klauer:
Would be cool to complete this

klauer:
Looks like I may have wasted time on #425 :sigh:

tacaswell:
wait, had 2 of these open?!

tacaswell:
oh, this is the static version.

klauer:
Yeah, confused me, too.

PR #394, opened 2018-11-18
FIX: only propose changing the alarm state if it is actually changed

tacaswell:
This prevents subscriptions for being processed for all signal that
share an alarm if the alarm state has not actually changed.


klauer:
Only failure is due to benchmarks - merging.

PR #405, opened 2018-12-06
FIX: strings invalid element count requested; max-length assumptions for unspecified values

### LINKED ISSUE #399: Server string handling + count
IOC with empty string:
--------------
```python
class EpicsDXPGroup(PVGroup):
    live_time_output = pvproperty_with_rbv(name='LiveTimeOutput', dtype=str)
```

Request
---------
`caget LiveTimeOutput` and even `caget -#0 LiveTimeOutput` fail with  `*** CA error Invalid element count requested`

`caget` and libca-based libraries such as pyepics fail with these requests.

Others
--------
There are other quirks with server string handling that need to be addressed as well, but I need to make a point of noting them down here...

klauer:
Empty arrays (primarily those of type char/string) behaved strangely with `caget`, sometimes spouting:

`*** CA error: invalid element count requested`

This appears to be due to `max_length` assumptions in `ChannelData` when the initial value is empty (empty string, empty list/array).

State before and after this PR's changes:
* Master: Unless specified otherwise, that the maximum length of data `ChannelData` will hold is the length of the initial value
* PR: If initial value is empty, assume scalar and set max_length=1.

Closes #399 (for now)

### DIFF
--- a/caproto/_data.py
+++ b/caproto/_data.py
@@ -280,7 +280,7 @@ class ChannelData:
 
         if self._max_length == 1:
             if is_array:
-                if value:
+                if len(value):
                     # scalar value in a list -> scalar value
                     return value[0]
                 else:
--- a/caproto/_data.py
+++ b/caproto/_data.py
@@ -212,7 +212,11 @@ class ChannelData:
         self.reported_record_type = reported_record_type
 
         if self._max_length is None:
-            self._max_length = self.calculate_length(value)
+            # Use the current length as maximum, if unspecified.
+            self._max_length = max(self.calculate_length(value), 1)
+            # It is possible to pass in a zero-length array to start with.
+            # However, it is not useful to have an empty value forever, so the
+            # minimum length here is required to be at least 1.
 
         value = self.preprocess_value(value)
         self._data = dict(value=value,
@@ -276,8 +280,12 @@ class ChannelData:
 
         if self._max_length == 1:
             if is_array:
-                # scalar value in a list -> scalar value
-                return value[0]
+                if value:
+                    # scalar value in a list -> scalar value
+                    return value[0]
+                else:
+                    raise CaprotoValueError(
+                        'Cannot set a scalar to an empty array')
         elif not is_array:
             # scalar value that should be in a list -> list
             return [value]
--- a/caproto/_circuit.py
+++ b/caproto/_circuit.py
@@ -245,7 +245,7 @@ class VirtualCircuit:
                     # has 0 payload, just drop it on the floor and
                     # move on.
                     return
-                # Otherwise, transmute the Command to a EventCancelRequest
+                # Otherwise, transmute the Command to a EventCancelResponse.
                 command = EventCancelResponse(command.data_type,
                                               ev_add.sid,
                                               command.subscriptionid,
@@ -386,7 +386,9 @@ class VirtualCircuit:
                 # send or received in the future are valid.
                 self.event_add_commands[command.subscriptionid] = command
             elif isinstance(command, EventCancelRequest):
-                # if we see a cancel request, note that so we know
+                # If we see a cancel request, note that so we know to interpret
+                # the next EventAddResponse with an empty payload as an
+                # EventCancelResponse.
                 self.event_cancel_commands[command.subscriptionid] = \
                     self.event_add_commands[command.subscriptionid]
             elif isinstance(command, EventCancelResponse):
--- a/caproto/_circuit.py
+++ b/caproto/_circuit.py
@@ -71,6 +71,8 @@ class VirtualCircuit:
         self._data = bytearray()
         self._ioids = {}  # map ioid to Channel
         self.event_add_commands = {}  # map subscriptionid to EventAdd command
+        # map subscriptionid to EventAdd command as we wait for them to die
+        self.event_cancel_commands = {}
         # There are only used by the convenience methods, to auto-generate ids.
         if our_role is CLIENT:
             self._channel_id_counter = ThreadsafeCounter(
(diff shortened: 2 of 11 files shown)


caproto/_data.py
```diff
@@ -276,8 +280,12 @@ def preprocess_value(self, value):
...
         if self._max_length == 1:
             if is_array:
-                # scalar value in a list -> scalar value
-                return value[0]
+                if len(value):
+                    # scalar value in a list -> scalar value
+                    return value[0]
+                else:
+                    raise CaprotoValueError(
+                        'Cannot set a scalar to an empty array')
```
tacaswell: Is this a protocol thing, an implementation detail of the c-version thing or a "This makes it better" thing?

tacaswell: This also seem weird that you can init with a 0 length array, but then the first time you set it becomes a scalar.   Is it ok to have a PV that we can not re-set (from the network) back to its initial value?

klauer: There is, admittedly, a bit of a loss of generality here. I don't personally think it's worthwhile to support the general case.

This change only affects the case where `max_length == 1` - i.e., you are working with a scalar value to start with. If we are to allow the empty array here, that would equate to passing in `None` to the putter functions that are expecting a scalar.

This is different from array values (`max_length > 1`) that can accept an array of length in `range(0, max_length + 1)`. The putter would be expected to handle arrays - including an empty array with the given `dtype`.

tacaswell: I'm sold on the loss of generality, but am concerned about the interaction with https://github.com/NSLS-II/caproto/pull/405/files#diff-bc1e46712779c588b17e8e8c6c01fc23R216 which demotes (promotes) 0 length init values to scalars rather an arrays with max length of 1 (is there a difference?)

caproto/_data.py
```diff
@@ -212,7 +212,11 @@ def __init__(self, *, alarm=None, value=None, timestamp=None,
         self.reported_record_type = reported_record_type
 
         if self._max_length is None:
-            self._max_length = self.calculate_length(value)
+            # Use the current length as maximum, if unspecified.
+            self._max_length = max(self.calculate_length(value), 1)
```
tacaswell: would it make more sense to blow up and refuse to to a 0 length input without also providing a max length?

danielballan: I would like that better.

caproto/_data.py
tacaswell: "I'm a scalar, and my functions expect that you'll pass in a scalar" also has an implicit `min_length == 1` which is different that every other `max_length` value...

klauer: Are you pushing for the inclusion of `min_length`, requiring our scalar-handling functions to allow `None`, or...?

tacaswell: mostly just thinking (mumbling) out loud.

klauer: I'd like to come to a consensus on these finer points before I make any additional changes...

tacaswell: Maybe tracking 'scalerness' as it's own state?

caproto/_data.py
danielballan: Meh, I have changed my mind about whether it is worth breaking currently-deployed IOCs over this. I think, "You didn't give me a length or a value, so I will assume this is a scalar," is acceptable.

danielballan: Maybe starting issuing warnings, though.

caproto/_data.py
danielballan: I think this is fine as is. Here is another protocol we like that has similar behavior:

```
def f(arg):
    # implicit min_length == 1
```

```
def f(*args):
    ...
```

tacaswell: "you can't have an array which is 0 or 1 long because ...... reasons" is also the sort of thing that if we came across else where we would get really annoyed at.

Also related, what happens if you try to set the setpoint of a motor to a zero length array?

This code is in the path run on init (around L218 in _data) so why is this not blowing up on empy initial values?

klauer: Let's back up a second here. When have you encountered an empty array over channel access in the wild?  If so, was it actually of any significance, uninitialized, or in error?

caproto/_data.py
klauer: Final answer? ;)

caproto/_data.py
tacaswell: Not sure I have seen one in the wild, but I think the protocol allows it...

"this PV is the list of photons I saw the last exposure" should be able to be empty? "list of files I have created from last reset"   I'm sure we could come up with some more stories with a bit of thought...

klauer: We can revisit this at a later date. For now I'll mark this as resolved in the interest of moving forward.

danielballan:
One test fails repeatably:

```
caproto/tests/test_threading_client.py::test_multiple_subscriptions_one_server[caproto]
```

with errors like

```
caproto._utils.RemoteProtocolError: The sid in EventCancelResponse(data_type=<ChannelType.CHAR: 4>, sid=1, subscriptionid=7, data_count=0) does not match the sid in in the original EventAddRequest for this subscriptionid, 9.
```

Investigating now, just taking notes as I go...

danielballan:
Rebased on master to remove all the `trio.Queue` deprecation warnings from the Travis logs. The error noted in my previous comment is reproducible locally.

danielballan:
Rebased on master again.

klauer:
Hmm, one test is failing consistently: `test_multiple_subscriptions_one_server[caproto]`

danielballan:
Saw the same a month ago (note comment above) so it is not due to anything recent. Apologies for the noise if that was already clear.

tacaswell:
powercycled to run against current master.

danielballan:
Notable that _one_ build passed. This must be an easy-to-hit-but-possible-to-miss race condition?

tacaswell:
and it looks like a bug in the threading client as we see it against both the caproto IOC and the epics-base IOC....

tacaswell:
turns out it was a bug in how we were handling an ambiguity / wart in the protocol that `EventCancelResponse` actually comes across the wire as `EventAddResponse` with 0 payload, we were identifying this and converting to `EventCancelResponse` at de-serialization time.  Instead we now respect the declared type at de-serialization time and keep track in `_circuit` if we have requested to cancel the subscription and only if so treat a 0 payload `EventAddResponse` as a `EventCancelResponse`

This means we now never actually see `EventCancelResponse` come across the wire, but that is _probably_ ok.

This also creates a weird race condition where 

-> cancel request
<- an update that really is an 0 length array
<- the cancel response

so instead of caching things forever, we just drop un-known `AddEventResponse` with 0 payload on the floor silently and move on with life.

This leaves a race condition in the protocol:

 -> cancel request
 <- actual update with 0 length array
 -> create new subscription that happens to re-use the subscription ID
 <-> hurray, new subscription
 <- cancel response from old subscription treated as 0 length update to new subscription

however I don't think there is anyway to prevent this (due to the glitch in the protocol), but in practical terms if you always increment the subscription ID it will be very rare to wrap all the way back around in the lag time between the last update of the old subscription and the cancel message coming in.

tacaswell:
also, no idea why this ever passed....

klauer:
Great work debugging that issue, @tacaswell and @danielballan!

PR #406, opened 2018-12-06
ENH: support get_with_metadata in pyepics_compat.PV

klauer:
For feature parity with upstream. 

Note that this has not yet been merged in pyepics, and we should wait until that happens first.


klauer:
OK, upstream support for this has been merged in pyepics. This is ready for review.

PR #412, opened 2018-12-18
API: Deprecate color_logs in favor of set_handler.

danielballan:
We copied the caproto logging utilities to bluesky and improved on it there,
thanks mostly to input from @mrakitin. This PR copies those improvement back
to caproto.


tacaswell:
The failure is a test IOC failing to start.

PR #413, opened 2018-12-19
DEBUG-log read/write request and responses always

danielballan:
This refines the work in #370 to ensure that _all_ read/write requests and responses, including those with `wait=False` and those initiated via a batched request, generate DEBUG messages in the coresponding PV/channel logger.

### DIFF
--- a/caproto/threading/client.py
+++ b/caproto/threading/client.py
@@ -1646,7 +1646,7 @@ class PV:
         # Stash the ioid to match the response to the request.
 
         event = threading.Event()
-        ioid_info = dict(event=event, pv=self)
+        ioid_info = dict(event=event, pv=self, request=command)
         if callback is not None:
             ioid_info['callback'] = callback
 
@@ -1722,7 +1722,7 @@ class PV:
                              data_type=data_type, data_count=data_count)
         if notify:
             event = threading.Event()
-            ioid_info = dict(event=event, pv=self)
+            ioid_info = dict(event=event, pv=self, request=command)
             if callback is not None:
                 ioid_info['callback'] = callback
 
--- a/caproto/threading/client.py
+++ b/caproto/threading/client.py
@@ -1312,13 +1312,16 @@ class VirtualCircuitManager:
                                   ca.WriteNotifyResponse)):
             ioid_info = self.ioids.pop(command.ioid)
             deadline = ioid_info['deadline']
+            pv = ioid_info['pv']
             if deadline is not None and time.monotonic() > deadline:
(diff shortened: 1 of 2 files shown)


tacaswell:
varying the contents of `ioid_info` is a bit sketch, but not the worst.  It might be worth sticking it on in all cases?

I wonder of moving to NamedTuple would be a useful thing to look at for performance tuning (later).

caproto/threading/client.py
```diff
@@ -1312,13 +1312,16 @@ def _process_command(self, command):
...
-                self.log.warn("Ignoring late response with ioid=%d because "
+                self.log.warn("Ignoring late response with ioid=%d regarding "
+                              "PV named %s because "
                               "it arrived %.3f seconds after the deadline "
                               "specified by the timeout.", command.ioid,
-                              time.monotonic() - deadline)
+                              pv.name, time.monotonic() - deadline)
                 return
 
+            pv.log.debug("%r: %r", pv.name, command)
```
mrakitin: Why `pv.log.debug(...)` is used here? Should it be `self.log.debug(...)` as above?

caproto/threading/client.py
```diff
@@ -2146,6 +2151,8 @@ def __exit__(self, exc_type, exc_value, traceback):
         deadline = time.monotonic() + timeout if timeout is not None else None
         for ioid_info in self._ioid_infos:
             ioid_info['deadline'] = deadline
+            pv = ioid_info['pv']
+            pv.log.debug("%r: %r", pv.name, ioid_info['request'])
```
mrakitin: See my comment above.

caproto/threading/client.py
tacaswell: @danielballan should verify, but this is so that you can look an the logging of just a single channel (aka PV) rather than all of the channels that are on the same circuit (aka IOC).

danielballan: Yes. The circuit logger (`self.log` here) shows everything that happens on a circuit, which is useful for debugging _caproto itself_ but often too much information for debugging hardware. The PV/channel loggers (`pv.log` here) are separate, and they let you focus the logging a little more.

In current master, we called `PV.log.debug` inside `PV.write()` and `PV.read()`. In this PR, we still `PV.log.debug`; we just do it from inside the circuit's receiving loop so that it also applies to batched requests and to writes that `PV.write()` is not waiting for.

mrakitin: @danielballan, @tacaswell, thank you for the detailed explanation here and in a private channel! That answers my question.

PR #415, opened 2019-01-09
ENH: Add 'attempts' and 'timeout' on Context, PV.

### LINKED ISSUE #409: threading client: add support for multiple attempts of PV.read/PV.write
I could see an argument for putting this in as a keyword argument to `read` and `write`. The default would be configurable by `PV` and also by `Context`, falling back in the same way that logging levels propagate. That is, if a `retry` isn't specified in a given call to `pv.read()`, fall back to `pv.retry`; if that is `None`, fall back to `Context.retry`. The factory default of `Context.retry` would be `1` (try once in total).

danielballan:
Closes #409

Tested interactively. Could use some real tests...gotta think about how to
mock this behavior effectively on a test server.

### DIFF
--- a/caproto/threading/client.py
+++ b/caproto/threading/client.py
@@ -1726,7 +1726,8 @@ class PV:
         return ioid_info['response']
 
     @ensure_connected
-    def write(self, data, *, wait=True, callback=None, timeout=None,
+    def write(self, data, *, wait=True, callback=None,
+              timeout=PV_DEFAULT_TIMEOUT,
               notify=None, data_type=None, data_count=None):
         """
         Write a new value. Optionally, request confirmation from the server.
--- a/caproto/threading/client.py
+++ b/caproto/threading/client.py
@@ -854,7 +854,7 @@ class Context:
         If None is specified, a fresh one is instantiated.
     timeout : float
         Number of seconds before a CaprotoTimeoutError is raised. This default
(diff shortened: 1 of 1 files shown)


danielballan:
P.S. This has been upgraded from "seems like a good idea" to "I have a pressing need for this at CMS to help debug something for Mark Rivers."

danielballan:
> I'm not sure I've seen a scenario where multiple read retries would be necessary

It will help us work around a stubborn race condition at CMS, where the server is evidently dropping random `ReadRequests`. Obviously the proper fix is on the server side, but this would still be useful to be able to dial up in the mean time and may be useful again.

Do you think we should also apply retries to writes? That seems potentially scarier, which is why I just applied it to reads.

klauer:
I think applying it to writes could be disastrous - definitely not a good idea.

danielballan:
Per discussion with @klauer, I have stripped out the "attempts" aspect of this and left only the timeout. So, the scope of this PR is now, "Instead of specifying the timeout per read/write operation, you can change the Context-wide default and also customize the default on a PV level."

danielballan:
2/4 builds passed, 2 with one random failure each, which were:

```
caproto/tests/test_examples.py::test_mocking_records
caproto/tests/test_pyepics_compat.py::test_get_with_metadata 
```

Merging.

PR #416, opened 2019-01-10
FIX: Do not warn if the same result arrives twice

danielballan:
It's only a problem if we see the same search result from different servers.
We should not warn about seeing two results from the same server.

### DIFF
--- a/caproto/threading/client.py
+++ b/caproto/threading/client.py
@@ -631,11 +631,12 @@ class SharedBroadcaster:
                             if name in self.search_results:
                                 accepted_address, _ = self.search_results[name]
                                 new_address = ca.extract_address(command)
-                                self.log.warning(
-                                    "PV %s with cid %d found on multiple servers. "
-                                    "Accepted address is %s:%d. "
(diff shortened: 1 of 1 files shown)


danielballan:
Tested on RPi on local network where I can reliably reproduce this issue. With this change, I see no warning.

mrakitin:
Good job! That should resolve the issue seen at QAS when we used caproto control layer.

tacaswell:
How do you produce the issue?

danielballan:
I haven't done anything special. It's just an RPi on the wifi running a caproto server, talking to my laptop on that same wifi network.

tacaswell:
I have a small worry that getting multiple searches back is evidence that we are also sending out too many search?

I see the logic for not re-trying searches that have taken a while to get back to us, but it also seems we should have some floor on how often we re-search for a PV?  Particularly at application / profile start up we are likely to be searching for many PVs and I can imagine we get through the "lets issue the searches for N PVs" faster than any server can respond which means we search for the first PV N times and end up sending out O(N**2) searches.

PR #220, opened 2018-05-28
CI: Pave the way for executing tests in parallel, along with other fixes

danielballan:
The xdist plugin seems to conflict with the benchmark plugin,
even when the benchmark plugin is disabled. Leaving this PR here
as a reminder to investigate later -- moving on to other stuff now.

PR #419, opened 2019-01-19
CI: also test py37

tacaswell:


danielballan:
Recall this was already discussed here: https://github.com/NSLS-II/caproto/pull/220#issuecomment-430860520

@klauer and I spend a lot of time wrestling with this CI and are reluctant to add builds, at least until we 
make more progress on the flaky threading tests (many of which are X-failed but really should not be).

danielballan:
To tiff on Brett Canon, if a PR is a "free puppy", a PR with an extra CI build is a "free pitbull with emotional damage."

tacaswell:
Sure, but  46f26e6df0663bf03b3dca98162b48df89210faf breaks asyncio servers on py37 but not py36.

tacaswell:
The failure seems to be a bug in cpython that was fixed in 3.7.1 (experimentally, having trouble finding any notes on what it could be).

tacaswell:
The issue that started me down this path turns out to have been an issue with how cpython is compiled (addressed by 077da6300cc930612e77f2d5685bd3a282f9d06a )

Closing in favor of #220 which also adds py37 tests.

PR #422, opened 2019-01-19
TST: Use the new Context timeout kwarg.

danielballan:
This might help with the several IOC example tests that sometimes randomly
fail on a timeout.


tacaswell:
From debugging in #220 I am becoming less convinced that we are CPU starved and more convinced that we have a subtle race condition / bug in the client's broadcaster logic.

klauer:
There's fairly good evidence to back @tacaswell's theory in ophyd's caproto tests. With the same IOCs and tests, pyepics passes consistently whereas caproto fails randomly.

danielballan:
I agree, closing.

PR #423, opened 2019-01-20
FIX: handle setting empty lists as values through pyepics compat

tacaswell:


danielballan:
Can you add a module docstring or comment at the top explaining why there is a separate file (2)?

PR #424, opened 2019-01-20
DOC: show vector PV and dtypes in simple IOC

tacaswell:


danielballan:
Good idea. Note to self: add some commits to update associated prose docs.

PR #425, opened 2019-01-20
ENH: a "clever" IOC example which dynamically creates pvproperties based on a formula

tacaswell:
The idea is that you can set a python expression to a PV and the
IOC will dynamically create / remove channels to have the PV names
match the variable in the expression.

Still needs some work (like being able to actually remove the old
channels!) and to sort out if this is just cute or actually a useful
idea.

### DIFF
--- a/caproto/ioc_examples/dynamic_calc.py
+++ b/caproto/ioc_examples/dynamic_calc.py
@@ -1,0 +1,58 @@
+#!/usr/bin/env python3
+from caproto.server import pvproperty, PVGroup, ioc_arg_parser, run
+from caproto import ChannelDouble
+
+import ast
+
+
+def extract_names(code):
+    m = ast.parse(code)
+    c = compile(code, '', 'eval')
+    # TODO also skip np namespace
+    return c, set(n.id for n in ast.walk(m)
+                  if isinstance(n, ast.Name))
+
+
+class SmartCalc(PVGroup):
+    formula = pvproperty(value='')
+    proc = pvproperty(value=0)
+    output = pvproperty(value=0.0)
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._pvdb = {}
+        self.out_db = {}
+        self.out_db.update(self.pvdb)
+
+    @formula.putter
+    async def formula(self, instance, value):
+        c, names = extract_names(value)
+
+        self.c = c
+        self.names = names
+
+        self._pvdb.clear()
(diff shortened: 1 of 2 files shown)


bl0x:
This is avery nice example, and without it I would have gotten stuck: Why is it, that one has to go through these hoops of maintaining a second out_db? Why isn't it possible to use pvproperty() or SubGroup() in the __init__ function or other functions of the top-level PVGroup? I'd like to dynamically generate or update the PVs on my IOC based on a general settings file (the IOC is only one part of the full program.). For example, I've tried to create a list of SubGroup's in a function that gets called before the IOC is started with the run() function, but that went completely sideways.

tacaswell:
@bl0x You want to look at the code in #393 .  In particular the code at https://github.com/NSLS-II/caproto/pull/393/files#diff-87472e30eeed1ebcb28f288d3a3cef9eR123 is creating a sub-class (of a subclass) of `PVGroup` which dynamically generates a static (for the life of the server) IOC.  I think this matches what you want (read the file at program start up, generate the required IOC, start it running and move on).

This example is doing something _much_ sketchier and adding (and eventually removing) channels from a running IOC on the fly.  I do not think this is something that can be done with epics-base so there is not a pressing need from a compatibility point of view to support this and it does sound like a nightmare to manage in production... As I said above,I am not yet sure if this example is useful or too cute for its own good.

klauer:
I think we need to be more clear on what IOCs we recommend for the general user versus power users like us who bend the system in fun-but-ugly ways most should not be copying...

I fully understand and respect that people now have access to creating IOCs in 100% Python and want dynamic _everything_. `PVGroup` was made to be largely static by design, along the lines of an ophyd `Device`. #429 tries to meet in the middle: "you have static portions, but can dynamically plug them together". I'm hopeful that will cover a good portion of these requests, since writing `ChannelData`-based IOCs isn't quite as fun.

 :x: ... if not, perhaps we need a third server API?

klauer:
I've had a bit of a change of heart with regard to these things as of late. An alternative that works _approximately_ within the bounds of what I consider acceptable is:
1. Have a base PVGroup for the IOC
2. Dynamically create a new PVGroup when necessary
3. `base_ioc.pvdb.update(**dynamic_group.pvdb)`

klauer:
Reworked this and put it into the appropriate spot. Will merge once tests pass.

klauer:
Lunchtime update - changed to never delete old items. Going to go with this version, as it doesn't require server refactoring.

tacaswell:
:tada:

PR #426, opened 2019-01-22
ENH: Add fine-grained loggers for search

ke-zhang-rd:
For issue [417](https://github.com/NSLS-II/caproto/issues/417#issuecomment-455804232)

### DIFF
--- a/caproto/_broadcaster.py
+++ b/caproto/_broadcaster.py
@@ -49,6 +49,7 @@ class Broadcaster:
         self._registered = False
         self._search_id_counter = itertools.count(0)
         self.log = logging.getLogger(f"caproto.bcast")
+        self.search_logger = logging.getLogger(f"caproto.search")
 
     def send(self, *commands):
         """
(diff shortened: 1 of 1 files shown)


caproto/_broadcaster.py
```diff
@@ -182,6 +183,9 @@ def search(self, name, *, cid=None):
             cid = self.new_search_id()
         commands = (VersionRequest(0, self.protocol_version),
                     SearchRequest(name, cid, self.protocol_version))
+        self.search_logger.debug("Search {name} and return {commands}")
```
danielballan: Instead of adding _additional_ log messages, I think it would be better to change the existing log messages to use more fine-grianed loggers. Most (all?) of the candidates are in `caproto/threading/client.py` where we use the `'caproto.bcast'` logger for many different things.

https://github.com/NSLS-II/caproto/blob/0f96128eb764054070b31e4a8c0e9b0e54407873/caproto/threading/client.py#L332

We should probably still use it for _some_ things, but anything related to a search request being sent or a search result being found should be directed to a `'caproto.serach.{name}'` logger.

danielballan:
Superceded by #427

PR #431, opened 2019-01-28
CI: maybe fix simulator fixture

tacaswell:
The test and the simulator fixture were getting different pvname
objects?


klauer:
Not really the case, at least locally. I believe the issue is the `xfail` environment check (currently, at least)... Closing for now.

PR #421, opened 2019-01-19
TST: Remove xfail on test_multithreaded_many_subscribe.

danielballan:
Let's see how this goes. Opening the PR to see what Travis thinks;
this has worked well *locally* for awhile now, but Travis is a different story.

PR #433, opened 2019-01-28
TST: re-add xfail on test_multithreaded_many_subscribe

klauer:


danielballan:
For anyone following along, we would soon like to remove this xfail, but there seems to be a subtle race condition or bug that presents itself often on Travis-CI but is so far impossible to reproduce locally. We will track progress on this bug in #421. Until then, we'll use xfail so that our CI is usable. #220 removed it prematurely.

PR #436, opened 2019-01-31
FIX: Add missing user_offset pvproperty to MotorFields

ronpandolfi:


tacaswell:
power cycled to run against current master which has the style fix.

PR #437, opened 2019-01-31
STY: flake8 found an over-indented method block

klauer:
New flake8 versions find over-indented method blocks now, apparently. 

Discovered in #436 travis results; blame indicates this has been there for a few months.


danielballan:
Merge at will.

PR #438, opened 2019-02-13
FIX: several things for pypy3

tacaswell:
No idea why this worked before under cpython...

### DIFF
--- a/caproto/_array_backend.py
+++ b/caproto/_array_backend.py
@@ -40,7 +40,6 @@ class Array(array.ArrayType):
             return Array(self.typecode, ret, endian=self.endian)
 
 
-
 type_map = {
     ChannelType.STRING: 'B',  # TO DO
     ChannelType.INT: 'h',
--- a/caproto/_array_backend.py
+++ b/caproto/_array_backend.py
@@ -1,4 +1,5 @@
 import array
+import copy
 import ctypes
 import sys
 from ._backend import Backend, register_backend, convert_values
@@ -31,6 +32,13 @@ class Array(array.ArrayType):
         self.endian = {'<': '>',
                        '>': '<'}[self.endian]
         super().byteswap()
+
+    if 'pypy' in sys.implementation.name:
+        def __deepcopy__(self, memo):
+            # this works around a bug in deepcopy on pypy
+            ret = copy.deepcopy(array.ArrayType(self.typecode, self))
+            return Array(self.typecode, ret, endian=self.endian)
+
 
 
 type_map = {
--- a/caproto/_array_backend.py
+++ b/caproto/_array_backend.py
@@ -13,7 +13,7 @@ default_endian = ('>' if sys.byteorder == 'big'
 
 class Array(array.ArrayType):
     'Simple array.array subclass which tracks endianness'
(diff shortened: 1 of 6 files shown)


klauer:
This appears to be a bug in pypy3. `_anonymous_` means that the fields of `'stamp'` should be directly accessible by the parent struct, without referencing `stamp.`. See: https://docs.python.org/3.7/library/ctypes.html#ctypes.Structure._anonymous_

That said, there's nothing lost here in making it pypy3-compatible, I suppose.

tacaswell:
ah, that makes sense!  Glad to have restored my sense of the world working...

tacaswell:
I tried to move the `_anonymous_` lines to each of the sub-classes and that seems to have broken it worse

```
==================================================================== ERRORS =====================================================================
_____________________________________________ ERROR collecting caproto/tests/test_serialization.py ______________________________________________
../../other_source/pypy/lib_pypy/_ctypes/basics.py:45: in from_param
    as_parameter = value._as_parameter_
E   AttributeError: 'TimeStamp' object has no attribute '_as_parameter_'

During handling of the above exception, another exception occurred:
caproto/tests/test_serialization.py:159: in <module>
    DBR_TIME_DOUBLE(1, 0, TimeStamp(3, 5))),
../../other_source/pypy/lib_pypy/_ctypes/structure.py:280: in __init__
    self.__setattr__(name, arg)
../../other_source/pypy/lib_pypy/_ctypes/structure.py:130: in __set__
    setattr(self.inside_anon_field.__get__(obj), self.name, value)
../../other_source/pypy/lib_pypy/_ctypes/structure.py:133: in __set__
    cobj = fieldtype.from_param(value)
../../other_source/pypy/lib_pypy/_ctypes/primitive.py:356: in from_param
    return super(SimpleType, self).from_param(value)
../../other_source/pypy/lib_pypy/_ctypes/basics.py:48: in from_param
    self.__name__, type(value).__name__))
E   TypeError: expected c_uint_be instance instead of TimeStamp

```

tacaswell:
This exposed 2 bugs in pypy (which I need to move over to them eventually)

Bug 1 is that ancestors are not consulted for `_anonymous_` https://bitbucket.org/pypy/pypy/src/271763ca6c718760604b8000ef18892758d39485/lib_pypy/_ctypes/structure.py#lines-196

Bug 2 is in how anonymous names are handled internally https://bitbucket.org/pypy/pypy/src/271763ca6c718760604b8000ef18892758d39485/lib_pypy/_ctypes/structure.py#lines-71:89  Looks like they are flattened and the anonymous field goes away (which seems wrong)

tacaswell:
There is an issue with numpy + pypy

```
(Pdb) buffers[2]
array([3.15])
(Pdb) memoryview(buffers[2]).cast('b')
*** TypeError: memoryview: cannot cast between two non-byte formats
```

which looks like either a numpy or a pypy bug, but no theories on that.

tacaswell:
the benchmarks require numpy and segfault when run...

tacaswell:
wait, travis is green ?!

klauer:
I re-ran 2 builds, that was it!

PR #441, opened 2019-02-16
MNT: more explicit socket closure

klauer:


klauer:
Travis and azure succeeded on the first try... hmm...

PR #442, opened 2019-02-16
REF: use ThreadsafeCounter; random start cid

klauer:


danielballan:
For the record, per an offline conversation, we consider the random start ID to be an extra layer of protection. It should not be necessary if everything is working correctly, but it significantly reduces the changes of the stars aligning with `cid` and `port`, causing interference and test failures.

The spec gives us room to do this. With regard to all the message identifiers, it says:

> The RECOMMENDED scheme for allocating these values is to create them sequentially starting at 0.

PR #444, opened 2019-02-17
FIX: leaking ioids

tacaswell:


caproto/_circuit.py
```diff
@@ -321,6 +321,13 @@ def _process_command(self, role, command):
                                       ClearChannelResponse)):
                 self.channels_sid.pop(chan.sid)
                 self.channels.pop(chan.cid)
+                to_remove = []
+                for k, v in self._ioids.items():
+                    if v is chan:
+                        to_remove.apend(k)
```
klauer: Typo
```suggestion
                        to_remove.append(k)
```

Or perhaps better:
```suggestion
                    to_remove = [k for k, v in self._ioids.items() if v is chan]
```

caproto/tests/test_core.py
```diff
@@ -367,3 +367,34 @@ def test_circuit_equality():
     b = ca.VirtualCircuit(ca.CLIENT, ('asdf', 1234), 1)
     c = ca.VirtualCircuit(ca.CLIENT, ('asdf', 1234), 2)
     assert a == b != c
+
+
+def test_ioid_exhastion(circuit_pair):
```
klauer:
```suggestion
def test_ioid_exhaustion(circuit_pair):
```

tacaswell:
It takes more than 100s to exhaust the ioids (I think in part because the logging is turned up to 11).

Is there a way to exempt this from the timeouts or the logging?

tacaswell:
ok, temporarily change the constants to be smaller :smiling_imp:

PR #447, opened 2019-02-18
ENH: caproto-shark

danielballan:
This adds a Python API and CLI for parsing pcap output (``tcpdump``) with
caproto. As in:

```bash
sudo tcpdump -U -w - port 50421 | caproto-shark
```

or

```python
from caproto.sync.shark import shark

parsed = shark(open('some_network_traffic.pcap', 'rb'))
```

There is, of course, already a wireshark plugin for CA. (We have been linking
to it in our documentation from the beginning.) There are situations where
using wireshark is a better choice. In other situations, it may be convenient
to analyze traffic in Python (making timing plots in matplotlib, etc.) and it
may be nice to have pcap analysis in caproto---batteries included.


doc/source/shark.rst
```diff
@@ -0,0 +1,128 @@
...
+
+   sudo tcpdump -w some_network_traffic.pcap port 5064
+
+Extract the information in Python using caproto:
+
+.. code-block:: python
+
+   from caproto.sync.shark import shark
+
+   file = open('some_network_traffic.pcap', 'rb')
```
tacaswell:
```suggestion
   with open('some_network_traffic.pcap', 'rb') as file:
       parsed = shark(file)
```

tacaswell: that is not going to apply well due to not removing the line below...

danielballan: I typed that first, but I think it could lead to someone trying to use parsed after the file had closed.

klauer:
Awesome! I've been wanting something like this for a while (and to make something like it for a while... you beat me to it :smile: ).

caproto/sync/shark.py
```diff
@@ -0,0 +1,280 @@
...
+            address = inet_ntoa(ip.src)
+            for command in read_datagram(transport.data, address):
+                yield SimpleNamespace(timestamp=timestamp,
+                                        ethernet=ethernet,
+                                        src=inet_ntoa(ip.src),
+                                        dst=inet_ntoa(ip.dst),
+                                        ip=ip,
+                                        transport=transport,
+                                        command=command)
+        else:
```
klauer: Should nuke the `else` clause here.

caproto/sync/shark.py
```diff
@@ -0,0 +1,280 @@
...
+                                        src=inet_ntoa(ip.src),
+                                        dst=inet_ntoa(ip.dst),
+                                        ip=ip,
+                                        transport=transport,
+                                        command=command)
+        elif isinstance(transport, UDP):
+            address = inet_ntoa(ip.src)
+            for command in read_datagram(transport.data, address):
+                yield SimpleNamespace(timestamp=timestamp,
+                                        ethernet=ethernet,
```
klauer: A bunch of flake8 over-hanging indents like this are causing travis to fail.

caproto/sync/shark.py
```diff
@@ -0,0 +1,280 @@
...
+
+def infer_command_class(header):
+    id_ = header.command
+    try:
+        return one_way_commands[id_]
+    except KeyError:
+        return sniffers[id_](header)
+
+
+def shark(file):
```
klauer: Could shark maybe made to capture non-CA packets but filter them out appropriately?

danielballan: I have been puzzling over the most reliable way to do that. Any ideas?

doc/source/shark.rst
danielballan: Nevermind, it's easy enough to guide people toward how to do this right. Change pushed.

caproto/sync/shark.py
klauer: Simplest would be to check each packet for something that looks like the header and try to deserialize it, ignoring ones that fail and adding them to an 'ignored' connections set (or something similar) for better performance.

Any alternative I can think of would be more complicated and fragile...

caproto/_commands.py
```diff
@@ -387,33 +387,31 @@ def read_from_bytestream(data, role):
 _commands = set()
 
 
-class _MetaDirectionalMessage(type):
-    # see what you've done, @tacaswell and @danielballan?
```
danielballan: It makes me sad to remove this line, but moving from metaclass to `__init_subclass__` is for the best, given that we have a reason to need class kwargs.

caproto/sync/shark.py
danielballan: Done. `tcpdump -w - | caproto-shark` (no source filtering) seems to work. We may discover edge cases as we use it -- things that look _enough_ like headers to trick us at first -- and the infrastructure is now in place to make validation smarter/stricter on the commands' respective `validate` meothods.

klauer:
One unrelated random test failure on azure; merging.

PR #456, opened 2019-03-09
Support for CLI output formatting arguments in 'get.py'

dmgav:
The proposed feature implements support for CLI arguments specifying output formatting for `caproto-get`:

- **floating point formats**: `-e -f -g -s -lx -lo -lb` 
- **integer numbers**: `-0x -0o -0b`

The arguments change output formatting whenever floating point or integer values are printed to `stdout`. If a format string is specified in the argument list (e.g. `--format="{pv_name} {response.data}"`), the changed formatting is applied to the field `{response.data}`. If the field specification differs (e.g. `{response.data[0]}`) then the default format or format supplied within `--format` string is used. For example, the argument sequence `-0b --format="{response.data[0]:X}"` will print integer as hex number, despite `-0b` argument that selects binary format. The formatting also works in `--wide` and `-terse` modes.   

EPICS caget accepts contradicting output formatting arguments and sets the format according to the last specified argument from the same group. Similar behavior is built into the proposed implementation. Parameters for float and int formats belong to different groups and are processed separately. For example, invocation of `caproto-get` (`simple.py` ioc):
```
caproto-get -f5 -0x -lx -0b simple:A simple:B
```
results in pv `simple:A` (int) to be printed in binary format and pv `simple:B` (float) to be rounded to the nearest int and printed in hex format.

**Custom separator**: `-F`. Sets custom separator for built-in or supplied format. The separator is inserted between fields and between array elements. Examples that set separator to `*` (consistent with the behavior of EPICS caget): `-F*`, `-F *`, `-F'*'`,`-F '*'`. EPICS caget accepts parameter -F without any value and sets separator to 's'. Such behavior is difficult to implement ('argparse' doesn't allow optional value), but it seems useless anyway. Also, EPICS caget accepts only single character separators, but the proposed implementation allows setting separator to a string of arbitrary length.
For example
``` 
caproto-get --a -F ' ** ' -e5 -0x simple:A simple:B simple:C
```
prints the pvs using standard `--wide` format string, setting separator between fields and elements of the array `simple:C` to `' ** '`. The floating point value of `simple:B` is printed in scientific format and int values of `simple:A` and `simple:B` are printed in hex format. 

The separator will also be inserted between fields if the format string is supplied with `--format` argument in case the format contains only spaces between the fields. For example if format string is specified as
```
-F ' ** ' --format="{pv_value}      {response.data[0]:X}" 
```
the separator will be inserted between two fields. If the definition is changed to
```
-F ' ** ' --format="{pv_value}      0x{response.data[0]:X}" 
```
by adding prefix `0x` before the second field, the separator will not be inserted. Note, that all spaces between `}` and `{` are replaced by the single separator: in the above example `}       {` is replaced by `} ** {`.

Added **support for standard EPICS type names** that start with `DBR_` (or `dbr_`). In EPICS `DBR_` prefix is optional and may be omitted. In the proposed implementation, if type name starts with `DBR_` (not case sensitive), the prefix is removed from the string before it is used for generating request. This feature improves compatibility with EPICS caget.

The **tests** were added for the proposed features. Unfortunately, the tests are basic and verify the exit codes returned by program in response to the sets of arguments, but not the formatting itself.

### DIFF
--- a/caproto/commandline/cli_print_formats.py
+++ b/caproto/commandline/cli_print_formats.py
@@ -21,10 +21,6 @@ class _DataPrintFormat:
         #                                      from the server as string
         self.no_brackets = False             # Print data as scalar (no brackets)
 
-    def is_set(self):
-        ''' Returns True if format string is not empty '''
-        return len(self.format) > 0
-
 
 def _clean_args_for_group(args_parsed=None, group_args=None, group_attrs=None):
     '''
@@ -211,7 +207,7 @@ def format_str_adjust(format_str=None, data_fmt=None):
     return format_str
 
 
-def print_response_data(data=None, data_fmt=None):
+def format_response_data(data=None, data_fmt=None):
     '''
     Prints data contained in iterable object 'data' to a string according to format specifications 'data_fmt'
     Returns a string containing printed data.
--- a/caproto/commandline/cli_print_formats.py
+++ b/caproto/commandline/cli_print_formats.py
@@ -1,0 +1,248 @@
+
+'''
+This module contains functions for formatting data output for CLI utilities
+(get.py and monitor.py). Format is selected based on command-line arguments.
+'''
+
+import numbers
+import sys
+import re
+
+
+class _DataPrintFormat:
+
+    def __init__(self):
+        self.format = ""
+        self.prefix = ""
+        self.separator = None
+        self.float_round = False    # floating point data value
+        #                             must be rounded to the nearest integer (long)
+        self.float_server_precision = False  # The floating point data value must be requested
+        #                                      from the server as string
+        self.no_brackets = False             # Print data as scalar (no brackets)
+
+    def is_set(self):
+        ''' Returns True if format string is not empty '''
+        return len(self.format) > 0
+
+
+def _clean_args_for_group(args_parsed=None, group_args=None, group_attrs=None):
+    '''
+    Cleans (sets to None) attributes of class 'args_parsed'. The only attribute that is
+    left corresponds to the argument that is specified last in the command line (sys.argv)
+
+    args_parsed - the class with attributes holding values of command line parameters
+    group_args - the list of the arguments in the group (ex. ["-0x", "-0o", "-0b"])
+    group_attrs - the list of corresponding attributes of 'args_parsed':
+                                                   ["int_0x", "int_0o", "int_0b"]
+    '''
+
+    if args_parsed is None or group_args is None or group_attrs is None:
+        return
+
+    if len(group_args) != len(group_attrs):
+        return
+
+    sa = sys.argv
+
+    arg_selected = None
+    for ag in reversed(sa):
+        for arg in group_args:
+            # Select appropriate regular expression for comparing arguments
+            reg_expr = "^{}$"  # Default is perfect match (typically for argument
+            #                    that starts with --)
+            # If the argument starts with '-', then match the beginning of the string
+            if len(ag) > 1 and ag[0] == '-' and ag[1] != '-':
+                reg_expr = "^{}.*"
+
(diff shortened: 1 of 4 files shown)


danielballan:
Wow, thanks for the contribution @dmgav! I look forward to reviewing this in detail in the next couple days, but in the meantime I want to express my gratitude for this useful, detail-oriented work.

klauer:
Seconding @danielballan's comment - thanks so much for the work here!

dmgav:
Thank you @danielballan and @klauer. I see that my submission does not pass AppVeyor test, but the reported errors are not related to the modules that I changed. Is the AppVeyor test up to date?

klauer:
AppVeyor can be safely ignored for now - it's not yet stable.

dmgav:
@danielballan, I believe that you did not have a chance to take a look at the code. I did a few more changes to the code and expanded support for the formatting arguments to `monitor.py`. I also moved the additional functions for processing of command arguments and output formatting to a separate module. As a result, the modified files `get.py` and `monitor.py` contain only minor changes compared with the original files. I also added a few tests for `monitor.py`.

I will upload the new version shortly, so that you could review all the changes at once when you have time.

caproto/commandline/cli_print_formats.py
```diff
@@ -0,0 +1,248 @@
...
+
+    # If a separator is specified (argument -F), then put the separators between each field
+    if data_fmt.separator is not None:
+        p = re.compile("} *{")
+        format_str = p.sub("}" + "{}".format(data_fmt.separator) + "{", format_str)
+
+    return format_str
+
+
+def print_response_data(data=None, data_fmt=None):
```
danielballan: Minor quibble: Reading the code where this is used, the name of this function momentarily misled me into thinking that it printed. Could this be named `format_response_data`?

caproto/commandline/get.py
```diff
@@ -77,7 +82,56 @@ def main():
...
+    fmt_group_int.add_argument('-0b', dest="int_0b", action="store_true",
+                               help=("Print as binary number"))
+
+    fmt_group_sep = parser.add_argument_group(title="Custom output field separator")
+    fmt_group_sep.add_argument(
+        '-F', type=str, metavar="<ofs>", action="store",
+        help=("Use <ofs> as an alternate output field separator (e.g. -F*, -F'*', -F '*', -F ' ** ')"))
+
     args = parser.parse_args()
+    # Remove contradicting format arguments. This function may be simply removed from code
```
danielballan: Agreed that this is messy but worth keeping for the sake of compatibility with the reference implementation.

caproto/commandline/cli_print_formats.py
```diff
@@ -0,0 +1,248 @@
...
+        self.separator = None
+        self.float_round = False    # floating point data value
+        #                             must be rounded to the nearest integer (long)
+        self.float_server_precision = False  # The floating point data value must be requested
+        #                                      from the server as string
+        self.no_brackets = False             # Print data as scalar (no brackets)
+
+    def is_set(self):
+        ''' Returns True if format string is not empty '''
+        return len(self.format) > 0
```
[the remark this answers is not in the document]
dmgav: I removed `is_set()` from code. It can easily be reimplemented if needed.

caproto/commandline/cli_print_formats.py
dmgav: You are right. I renamed the function  to `format_response_data`.

dmgav:
Thank you for reviewing the code. I implemented the proposed minor changes and uploaded the corrected version.

danielballan:
Thanks for your patience with my slow review and for the fast turn-around. Great, well-scoped contribution!

PR #458, opened 2019-03-26
FIX: Be more paranoid about non-IP traffic.

danielballan:


klauer:
```sh
$ flake8 .
./caproto/sync/shark.py:7:1: F401 'dpkt.ip6.IP6' imported but unused
flake8 FAILED!
```

Issue #452, opened 2019-03-05
Pluggable search


danielballan:
Clients should be able to consult a mapping before initiating UDP broadcast. This mapping could be backed by:

* aplain text file of PV names and addresses
* etcd
* channel finder?

Any of these is likely to be significantly faster than UDP broadcast. If that mapping fails to resolve successfully, UDP will still be available as a fallback. On the server side, servers need to be able to update this mapping at startup, shutdown, and at arbitrary points during their runtime (see new dynamic IOC exmaples).

klauer:
Related: we should determine how we're going to keep these "`CA++`"-style additions cleanly separate - but yet still accessible - from the base CA-only implementation.

danielballan:
Yes. I think that question is important for #453. For this, I think clients are always allowed to initiate the TCP phase with a server at some address they happen to know, so this it just surfacing API to make this more obvious. I wouldn't say that this even rises to the level of an extension to the protocol.

tacaswell:
I would weakly argue that this is not an extension of the protocol because clients are not _obligated_ to search before connecting.  We already have a notion of caching the search results with in the client, this is just pushing that cache out / making it fancier.

I would also add using a redis server and consulting it first before doing the broadcast should also be added to the list (expiration via https://redis.io/commands/expire like we do now in-process).

klauer:
Regardless of how you classify it - these fun enhancements/additions should be very optional and out-of-the-way.

PR #466, opened 2019-04-02
SearchResults container idea

klauer:
Here's a thought (a currently non-working one):
Break out search result storage and thread-safe updating into its own class

Advantages:
* Easier to keep locking logic in your head
* Locking can be done separately from broadcaster class state
* Could write some tests directly against the search results (maybe even fuzzing); to verify its functionality
*  (@danielballan noted below) Potential re-use in all clients
*  (@danielballan noted below) Much more easily tweakable, allowing for further customization as in #452

In my usual poor form, this got confused with another attempt at removing an awkward loop I had previously tagged with a `TODO`. For the meat of this PR, look at the latest commit.

### DIFF
--- a/caproto/threading/client.py
+++ b/caproto/threading/client.py
@@ -306,6 +306,23 @@ class SearchResults:
 
         1. `SearchResults()[name] -> [time, address]`
         2. `name in SearchResults()`
+
+    Attributes
+    ----------
+    name_to_addr : dict
+        Holds search results.
+        Maps name -> (time, address)
+    addr_to_name : dict
+        Holds search results.
+        Maps address -> {name1, name2, ...}
+    _unanswered_searches : dict
+        Holds pending searches
+        Maps search_id -> [name, results_queue, retirement_deadline]
+    _lock : threading.RLock
+        Lock for internal updates to SearchResults status
+    _search_id_counter : ThreadsafeCounter
+        Counter for new searches. This will be kept in sync with
+        _unanswered_searches such that there is no overlap in keys.
     '''
 
     def __init__(self):
--- a/caproto/threading/client.py
+++ b/caproto/threading/client.py
@@ -786,7 +786,6 @@ class SharedBroadcaster:
                 continue
 
             queues.clear()
-            now = time.monotonic()
             for command in commands:
                 if isinstance(command, ca.Beacon):
                     now = time.monotonic()
--- a/caproto/threading/client.py
+++ b/caproto/threading/client.py
@@ -1176,11 +1176,10 @@ class Context:
             pvs.append(pv)
             name, _ = key
             names.append(name)
-            # If there is a cached search result for this name, expire it.
 
-            # TODO: search lock
-            # self.broadcaster.search_results.mark_channel_disconnected(
-            #     name, pv.address)
+            # If there is a cached search result for this name, expire it.
+            self.broadcaster.search_results.mark_channel_disconnected(
+                name, pv.channel.circuit.address)
 
             with self.pv_cache_lock:
                 self.pvs_needing_circuits[name].add(pv)
--- a/caproto/threading/client.py
+++ b/caproto/threading/client.py
@@ -464,7 +464,7 @@ class SharedBroadcaster:
             # TODO this is very inefficient
             for context in self.listeners:
                 for addr, cm in context.circuit_managers.items():
-                    if cm.connected and name in cm.all_created_pvnames:
+                    if cm.connected and cm.pvname_to_channels[name]:
                         # A valid connection exists in one of our clients, so
                         # ignore the stale result status
                         with self._search_lock:
@@ -1234,7 +1234,7 @@ class VirtualCircuitManager:
     """
     __slots__ = ('context', 'circuit', 'channels', 'ioids', '_ioid_counter',
                  'subscriptions', '_ready', 'log',
-                 'socket', 'selector', 'pvs', 'all_created_pvnames',
+                 'socket', 'selector', 'pvs', 'pvname_to_channels',
                  'dead', 'process_queue', 'processing',
                  '_subscriptionid_counter', 'user_callback_executor',
                  'last_tcp_receipt', '__weakref__')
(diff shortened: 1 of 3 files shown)


danielballan:
Innnntersting. In early iterations of caproto's core and initial clients (which were the async ones) I recall waffling on how much of the search logic should be held by the core. In the end we settled on, "basically none of it." But it does seem likely to be reused across clients. And as complex as it is now, it will only become more so if we incorporate #452. Could it be useful to put something like `SearchResults` in `caproto._broadcaster` and put a thread-locking subclass in `caproto.threading.client`?

klauer:
Yes to all of the above, @danielballan!

The relevance to the non-UDP search option briefly crossed my mind while I dreamt about this on the plane, too.

And as you noted, since we need to get the other async clients up-to-speed as well at some point, having this accessible to them as well would make development much easier. To your suggestion - do you think there's any harm in just using the `threading.Lock`-based version in all clients?

danielballan:
> To your suggestion - do you think there's any harm in just using the `threading.Lock`-based version in all clients?

I think it's reasonable to just use `threading.Lock`. If we find a good reason to add another layer we can revisit the question, but "purity" is not a good enough reason. (My view has moved on this since 2016.)

tacaswell:
I'm on board with this :100: %

I'm mixed on the sync vs async split.  There is something to be said for "threads may only communicate via Queues" (zmq actually makes this pitch in their docs!) .  That is basically the pattern we use everywhere else and is very much inline with the sans-io philosophy (inter-thread / task communication is I/O!).

klauer:
(This is not ready by any stretch. Only marking this as "ready" to get CI, which apparently does not run on draft PRs.)

klauer:
I wasn't expecting this to pass - perhaps it's in better shape than I had thought.

With this `SearchResults` idea, I also think there is less benefit to the `SharedBroadcaster` concept. The main thing we did not want was multiple contexts all performing the same searches and spamming the network needlessly.

If all `Broadcaster` instances share a single `SearchResults` instance, perhaps `SharedBroadcaster` can return to being just `Broadcaster`, each associated with a single context?

caproto/threading/client.py
```diff
@@ -300,6 +309,240 @@ def __call__(self):
...
+        '''
+        retirement_deadline = time.monotonic() + SEARCH_RETIREMENT_AGE
+        for item in self._unanswered_searches.values():
+            item[-1] = retirement_deadline
+
+    @property
+    @_locked
+    def unanswered_searches(self):
+        'All unanswered searches'
+        return dict(self._unanswered_searches)
```
danielballan: To save from copying on access, it might be worth stealing this `Frozen` class from xarray: https://github.com/pydata/xarray/blob/3fb22cbf86e49be4a444fa062350bf7570abd6da/xarray/core/utils.py#L333-L356

caproto/threading/client.py
```diff
@@ -325,18 +568,14 @@ def __init__(self, *, registration_retry_time=10.0):
         # PVs (via Context.get_pvs).
         self._search_now = threading.Event()
 
-        self.search_results = {}  # map name to (time, address)
-        # map search id (cid) to [name, queue, last_search_time, retirement_deadline]
-        self.unanswered_searches = {}
+        self.results = SearchResults()
```
danielballan: Is this superfluous?

caproto/threading/client.py
```diff
@@ -300,6 +309,240 @@ def __call__(self):
...
+            except KeyError:
+                needs_search.append(name)
+            else:
+                use_cached_search[address].append(name)
+
+        return use_cached_search, needs_search
+
+
 class SharedBroadcaster:
     def __init__(self, *, registration_retry_time=10.0):
```
danielballan: Per your suggestion about moving to one-broadcaster-per-content, I guess `search_results` should an optional parameter to `Broadcaster.__init__`.

caproto/threading/client.py
klauer: Yes, good catch.

caproto/threading/client.py
klauer: I think we can just remove this altogether. This is a remnant of previous, less-clean abstraction that required `Broadcaster` to reach into `SearchResults.unanswered_searches`.

caproto/threading/client.py
klauer: I think so, but we may want to move that refactor to be done later. I'd like `SearchResults` to be done right (it still is not), and then move toward removing `SharedBroadcaster` separately.

caproto/threading/client.py
```diff
@@ -300,6 +317,235 @@ def __call__(self):
...
+        self.addr_to_names.clear()
+        self._unanswered_searches.clear()
+        self._searches_by_name.clear()
+        self._searches.clear()
+
+    @_locked
+    def mark_server_alive(self, addr):
+        'Beacon from a server received'
+        for name in self.addr_to_names[addr]:
+            self.name_to_addrs[name][addr] = time.monotonic()
```
tacaswell: grab the time once before the loop so they all have the same time?

danielballan:
I addressed the review comments to nudge this along. Seems like it may be close to merge-able.

klauer:
It's not _too far_ off. More interactive testing should be done, if possible, prior to merging.

Also - for what it's worth, I'm not tied to any of the methods/names/etc. presented here. If anyone has an idea for improvement, by all means let me/us know.

caproto/threading/client.py
```diff
@@ -300,6 +317,236 @@ def __call__(self):
...
+
+        1. `SearchResults()[name] -> [time, address]`
+        2. `name in SearchResults()`
+
+    Attributes
+    ----------
+    name_to_addrs : dict
+        Holds search results.
+        Maps name -> [(address, name, time), ...]
+    addr_to_name : dict
```
klauer: Docstring needs updating for new attributes

caproto/threading/client.py
```diff
@@ -300,6 +317,236 @@ def __call__(self):
...
+        Holds search results.
+        Maps address -> {name1, name2, ...}
+    _unanswered_searches : dict
+        Holds pending searches
+        Maps search_id -> [name, results_queue, retirement_deadline]
```
klauer: `[name, results_queue, resend_deadline, retirement_deadline]`

danielballan:
Yeah, I'm going to take it for a spin with my household IOCs. Would also be nice to get a test at a beamline.

klauer:
A good idea, but not going anywhere. Closing.

PR #470, opened 2019-04-13
FIX: only re-send search request if we have not sent it recently

tacaswell:
This fixes a bug where if the user searches for many PVs
simultaneously (for example, when connecting a complex ophyd device)
the early requests will be reset on each subsequent request an
interval much shorter than MIN_RETRY_SEARCHES_INTERVAL.

### DIFF
--- a/caproto/threading/client.py
+++ b/caproto/threading/client.py
@@ -419,7 +419,8 @@ class SharedBroadcaster:
             self.udp_sock = None
 
         self._close_event.set()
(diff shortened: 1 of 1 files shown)


danielballan:
Any idea of this is what epics-base does? I am more merging in any case because this is an important fix to my work in 0.2.3 until we have time to get all the details right.

tacaswell:
nope, did not fix race condition (only 1/8 passed) :(

PR #472, opened 2019-04-15
MNT: make enum field on scalars_and_arrays mock a bi record

tacaswell:
This was sitting in a stash on my laptop, I think this will make the
example IOC play nicer with viewers.


klauer:
I think we should be linking the enum strings of the `ChannelEnum` with the `ONAM`/`ZNAM` of `BiFields`/`BoFields` by default:

https://github.com/NSLS-II/caproto/blob/bd21dc83eae7c50aa10253d5f6d60f7d93df191a/caproto/server/records.py#L1257-L1260

klauer:
Followed up with #485

PR #474, opened 2019-04-25
ENH: Add pvs whose prefex=complex: for log filter wildcard test

ke-zhang-rd:
Run commands below on different terminal.

`python3 -m caproto.ioc_examples.complex --list-pvs -vvvv`
`python3 -m caproto.ioc_examples.simple --list-pvs -vvvv`

### DIFF
--- a/caproto/ioc_examples/complex.py
+++ b/caproto/ioc_examples/complex.py
@@ -1,0 +1,29 @@
+#!/usr/bin/env python3
+from caproto.server import pvproperty, PVGroup, ioc_arg_parser, run
+from textwrap import dedent
+
+
+class ComplexIOC(PVGroup):
+    """
+    An IOC with three uncoupled read/writable PVs
+
+    Scalar PVs
+    ----------
+    A (int)
+    B (float)
+
+    Vectors PVs
+    -----------
+    C (vector of int)
+    """
+    A = pvproperty(value=1)
+    B = pvproperty(value=2.0)
+    C = pvproperty(value=[1, 2, 3])
+
+
+if __name__ == '__main__':
+    ioc_options, run_options = ioc_arg_parser(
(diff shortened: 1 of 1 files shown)


ke-zhang-rd:
I was trying to write `complex:*` and `simple:*` inside same file and load it by one `run`. But it isn't trivial for me how to do that.

danielballan:
Good question. I think the right approach here is to put all the PVs you need into one IOC. The existing [`subgroups` IOC example](https://github.com/NSLS-II/caproto/blob/master/caproto/ioc_examples/subgroups.py) might be suitable for your needs.

```
$ python -m caproto.ioc_examples.subgroups --list-pvs
random using the descriptor getter is: <caproto.server.server.PvpropertyInteger object at 0x7fb1b30f99b0>
subgroup4 is: <__main__.MyPVGroup.group4.subgroup4 object at 0x7fb1b30f96a0>
subgroup4.random is: <caproto.server.server.PvpropertyInteger object at 0x7fb1b30f9860>
[I 09:05:15.942 server:158] Asyncio server starting up...
[I 09:05:15.942 server:170] Listening on 0.0.0.0:5064
[I 09:05:15.942 server:253] Server startup complete.
[I 09:05:15.942 server:255] PVs available:
    subgroups:random
    subgroups:RECORD_LIKE1.RTYP
    subgroups:RECORD_LIKE1.VAL
    subgroups:RECORD_LIKE1.DESC
    subgroups:recordlike2.RTYP
    subgroups:recordlike2.VAL
    subgroups:recordlike2.DESC
    subgroups:group1:random
    subgroups:group2-random
    subgroups:group3_prefix:random
    subgroups:group4:subgroup4:random
```

And if that doesn't have the right variety of PV names for testing, some variation on that should get you want you want.

PR #475, opened 2019-04-26
MNT: modify help docstring

ke-zhang-rd:
asyncio as default in help docs


danielballan:
Looks good. Test failures are known random failures, obviously not related to this docs change. Thanks for fixing this oversight!

PR #480, opened 2019-05-13
CI: Fix typo in doctr deploy command.

danielballan:
Correction to #479. Once it was merged and tried to publish the docs, we can
see that [it failed](https://travis-ci.org/caproto/caproto/jobs/530885729)
becauasethe token was (correctly) set up for caproto.github.io
but the command to push tries (incorrectly) to publish to bluesky.github.io.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -108,5 +108,5 @@ script:
   # Upload the docs to gh-pages.
   - if [[ $PUBLISH_DOCS ]]; then
         doctr deploy --deploy-repo NSLS-II/NSLS-II.github.io --deploy-branch-name master caproto;
-        doctr deploy --deploy-repo bluesky/bluesky.github.io --deploy-branch-name master caproto;
+        doctr deploy --deploy-repo caproto/caproto.github.io --deploy-branch-name master caproto;
     fi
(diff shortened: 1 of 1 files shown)


danielballan:
Failing on some bad interaction between doctr and submodules I think. No idea why this didn't come up before.

https://travis-ci.org/caproto/caproto/jobs/531866387#L7026

klauer:
Is this OK to go in? I only ask as it's been quite a while...

danielballan:
Unfortunately no. I haven't found a chance to dig into whatever bad interaction is happening between doctr and submodules.

danielballan:
@mrakitin I believe the problem is related to a bad interaction between doctor and submodules, so none of our other repos would be affected. The best way to fix this for all repos is to apply a patch to doctr itself but I don’t understand the problem yet well enough to knelt what that patch would be.

Note well that this would-be workaround is not actually working yet, and this PR is not reddy to merge. I hope I’m close though....

danielballan:
Solved! The issue was the caproto.github.io had never been initialized. It was at this phase of repo setup:

image (8)

Pushing an initial commit there made this work properly.

PR #483, opened 2019-05-13
ENH: pvproperty doc kwarg becomes DESC

klauer:
Something like this:
```python
    heartbeat = pvproperty(value=0, mock_record='bo', doc='The heartbeat signal')
```

Becomes:
```bash
$ caget -S heartbeat.RTYP
heartbeat.RTYP bo

$ caget -S heartbeat.DESC         # <----
heartbeat.DESC The heartbeat signal
```


klauer:
I may have done this reverse of how it should be - the fields classes should probably be linking to the parent description instead, as in how others are done. Will follow up later.

klauer:
Reworked this to be similar to #485, and tweaked a test to verify reading/writing of `DESC`.

PR #484, opened 2019-05-13
ENH: allow init kwargs to pass through to SubGroups

klauer:
TODO/refactor: `valid_init_kw` is copy/pasted in a few places here

### DIFF
--- a/caproto/server/server.py
+++ b/caproto/server/server.py
@@ -716,7 +716,7 @@ class SubGroup:
             self.attr_separator = attr_separator
         self.base = (PVGroup, ) if base is None else base
         self.__doc__ = doc
-        self.init_kwargs = dict(init_kwargs) if init_kwargs else {}
+        self.init_kwargs = init_kwargs
         # Set last with setter
         self.group = group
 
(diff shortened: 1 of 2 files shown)


caproto/server/server.py
```diff
@@ -697,6 +698,20 @@ def group(self, group):
             assert inspect.isclass(group), 'Group should be dict or SubGroup'
             assert issubclass(group, PVGroup)
             self.group_cls = group
+            valid_init_kw = {
```
tacaswell: This means nothing in the chain can use `**kwargs` in a way that is not just pass-through.

Do we have many cases where we have significantly deferred instantiation of the instances?  I think in a vast majority of cases we instantiate all of the instances to build the pvdb to pass to the server so even without this check, we will get failures at server start time.

To get useful tracebacks it may be better to wrap the init call in L1216 (old code) with something like

```python
try:
    inst = subgroup_cls(...)
except Exception as e:
    raise RuntimeError(f'informative message') from e
``` 
?

klauer: It's been a couple of months here, so I don't really recall the details entirely. I think my logic was that the usage of `kwargs` already is so very confusing that the extra layer of annoying exceptions just made sense.

I'll close this and reopen another PR at a later date.

PR #485, opened 2019-05-14
ENH: link ZNAM and ONAM to the parent enum_strings

klauer:
Specify an enum in a `pvproperty`, and `ZNAM` and `ONAM` for mocked `bi`/`bo` records will be automatically linked.

Example IOC + test added.


klauer:
Rebased on master and added test.

PR #488, opened 2019-05-22
FIX: Handle case where filter_ is None or empty.

danielballan:
Attempt to fix a bug reported on the Nikea Slack where `filter_` is `None`:

```python
        modifiers = parse_record_field(name).modifiers
        if modifiers is not None:
            self.channel_filter = parse_channel_filter(modifiers.filter_)
```


tcornall:
Thanks Daniel.
This fixed that problem.

PR #492, opened 2019-06-07
FIX: EPICS_CA_REPEATER_PORT needs to be an integer

klauer:
`caproto-get` fails confusingly with the following without this fix:
```
$ caproto-get VMTEST:m1
an integer is required (got type str)
```

There may be other cases where we aren't changing these to `int()`...

### DIFF
--- a/caproto/sync/client.py
+++ b/caproto/sync/client.py
@@ -5,7 +5,6 @@
 import inspect
 import getpass
 import logging
-import os
 import selectors
 import socket
 import threading  # just to make callback processing thread-safe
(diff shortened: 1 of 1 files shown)


klauer:
(Force-pushed as I distractedly pushed a not-so-correct first attempt)

danielballan:
Pushed commit that should fix CI. Also grepped for all usages of `os.environ` and it looks like this was the only `int` problem.

danielballan:
Worst build matrix I've seen in awhile:

image

klauer:
And very slow, at that...

danielballan:
Re-running Travis....

klauer:
This should be ready to go, as failures are unrelated.

PR #494, opened 2019-06-14
ENH: add an example that publishes a gamepad input to the network

tacaswell:
```
caproto-monitor    gp:a    gp:b    gp:x    gp:y    gp:lb    gp:rb    gp:lx    gp:ly    gp:rx    gp:ry    gp:dx    gp:dy    gp:lt    gp:rt    gp:sel    gp:back    gp:alive gp:ld gp:rd

```

Attn @oksanagit

Yesterday Oksana and I were talking about an IOC for controlling
motors via a game pad.  This IOC gets as far as publishing the state
of the gamepad to EPICS.

### DIFF
--- a/caproto/ioc_examples/gamepad.py
+++ b/caproto/ioc_examples/gamepad.py
@@ -1,0 +1,113 @@
+#!/usr/bin/env python3
+"""
+This example requires a gamepad and the Python library evdev.
+
+This only works on linux (as it uses the linux kernel input events).
+"""
+from caproto.server import pvproperty, PVGroup, template_arg_parser, run
+from evdev import InputDevice
+
+
+class GampadIOC(PVGroup):
+    def __init__(self, event_path, **kwargs):
+        super().__init__(**kwargs)
+        self._event_path = event_path
+
+    # normal buttons
+    a = pvproperty(value=0, max_length=1)
+    b = pvproperty(value=0, max_length=1)
+    x = pvproperty(value=0, max_length=1)
+    y = pvproperty(value=0, max_length=1)
+    lb = pvproperty(value=0, max_length=1)
+    rb = pvproperty(value=0, max_length=1)
+
+    # joysticks
+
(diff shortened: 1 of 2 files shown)


caproto/ioc_examples/gamepad.py
```diff
@@ -0,0 +1,119 @@
...
+
+    # joysticks
+
+    # left joystick
+    lx = pvproperty(value=0, max_length=1)
+    ly = pvproperty(value=0, max_length=1)
+    ld = pvproperty(value=0, max_length=1)
+
+    # right joystick, always analog
+    rx = pvproperty(value=0, max_length=1)
```
klauer: `max_length` is assumed 1 if you pass `pvproperty` a scalar

Also, if this is an analog, it should be `0.0` instead of `0` to get the logic to pick `float` instead of `int`.

tacaswell: It is analog, but it seems to be ints in the range [-32k, 32k] vs the 'digital' which seem to be [-1, 0, 1]

klauer: Thinking a bit more about this - unless we can be certain that these inputs will always be 16-bit (which is -32k to 32k), it'd be preferable to normalize those values somehow.

klauer:
It'd also be possible to use a different library to make this multiplatform - perhaps pysdl2, pygame, or one of the many others out there. I think Linux-only is fine for now, though.

Think you're mostly done with the example, @tacaswell?

danielballan:
@tacaswell?

tacaswell:
I would prefer to leave this as is.

klauer:
Moved to hardware (and OS-) specific example directory.

klauer:
(Unrelated test failure, merging)

_Finally_, with this PR you can play games on the beamline without requiring a free USB port on the hutch computer (**). 

(**) Some assembly required.

PR #495, opened 2019-07-03
Add an option to perform a write without running the user's put method.

mattgibbs:
This PR adds a new option called 'with_callback' to `ChannelData.write()` that lets you choose whether to run the user-defined put method.

### DIFF
--- a/caproto/_data.py
+++ b/caproto/_data.py
@@ -477,11 +477,14 @@ class ChannelData:
 
         return (await self.write(value, flags=flags, **metadata_dict))
 
-    async def write(self, value, *, flags=0, **metadata):
+    async def write(self, value, *, flags=0, with_callback=True, **metadata):
         '''Set data from native Python types'''
         try:
             value = self.preprocess_value(value)
-            modified_value = await self.verify_value(value)
+            if with_callback:
+                modified_value = await self.verify_value(value)
+            else:
+                modified_value = None
         except GeneratorExit:
             raise
         except Exception:
--- a/caproto/_data.py
+++ b/caproto/_data.py
@@ -477,11 +477,11 @@ class ChannelData:
 
         return (await self.write(value, flags=flags, **metadata_dict))
 
-    async def write(self, value, *, flags=0, with_callback=True, **metadata):
+    async def write(self, value, *, flags=0, verify_value=True, **metadata):
         '''Set data from native Python types'''
         try:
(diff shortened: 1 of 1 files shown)


tacaswell:
This seems reasonable, but I am not sure about the name of the parameter.  I know that it is logically a user callback, but the name miss-match between `with_callback` and it skipping calling `verify_value` is a bit jarring.

danielballan:
Maybe flip the sign (`False` by default) and give it name like `skip_verify` or `skip_user_put` or `skip_user_processing`? I recall @mattgibbs help an informal poll on Slack and I failed to weigh in then.

klauer:
Maybe even just call the kwarg `verify_value`? It has a boolean-sounding name, and that's the method it'll call anyway.

danielballan:
And---belatedly---thanks for working on this @mattgibbs! Always happy to see a new contribution to caproto. :-D

mattgibbs:
I'll rename `with_callback` to `verify_value`.

I want to note that the naming isn't consistent within caproto: if I make a channel via the `@pvproperty` decorator, I define a "putter", not a "verifier".  Renaming `ChannelData.verify_value` to `ChannelData.putter` would be very nice, if not for the whole 'breaking change' thing.  In practice, I usually use `verify_value` all the time, to do all kinds of crazy stuff, but have never once used it to verify a value.

klauer:
@mattgibbs I can't really defend the current state of the codebase, other than saying it's the not-so-unusual case of "our weekend/off-hours prototype project became production".

In any case, the original goal of `verify_value` was not to do crazy things (although we've all been guilty of that through `pvproperty.putter`). Its purpose should remain as checking one value and return a potentially modified/fixed version of it.

Perhaps a fairly easy change would be to add a `putter` method on `ChannelData` that continues to call `verify_value`. It may be possible to weave this in such a way that could be backward-compatible, but don't quote me on that just yet.

mattgibbs:
Attn @danielballan, I'm bumping this - I think it is ready to merge.

danielballan:
Thanks for the bump. Looks good, though _all_ the Travis tests failed. The problem may be due to bit rot and is likely unrelated to your PR, but I'd prefer to investigate that before we merge.

danielballan:
Rebased.

klauer:
Looking past my ignored reply above, I guess we can just go with this.

PR #498, opened 2019-07-14
caproto-put using a file as the input value

gwbischof:


klauer:
Seems like it could be useful. Maybe for these `caproto-put` flags that don't exist in epics-base `caput`, we require the full `--file` and don't add the single-letter `-f`?

danielballan:
@gwbischof I like @klauer's suggestion because it should lower the chances of conflicting with parameters that may be added to the official `ca*` utilities in the future. Mind making the modification? Then I think this is good to merge---a handy addition to caproto.

caproto/commandline/put.py
```diff
@@ -66,6 +66,8 @@ def main():
     parser.add_argument('--array', '-a', action='store_true',
                         help=("Interprets `data` as an array, delimited by "
                               "space"))
+    parser.add_argument('--file', '-f', action='store_true',
```
klauer:
```suggestion
    parser.add_argument('--file', action='store_true',
```

caproto/commandline/put.py
```diff
@@ -85,7 +87,11 @@ def main():
             set_handler(color=not args.no_color, level='DEBUG')
     logger = logging.LoggerAdapter(logging.getLogger('caproto.ch'), {'pv': args.pv_name})
 
-    if args.array:
+    if args.file:
+        print(args.file, args.data)
```
klauer:
```suggestion
```

danielballan:
Reconciling this with #506 needs some thought. The user might want `-S --file <filename>`. As written, using `-S` causes `--file` to be silently ignored.

One solution that occurs to me is to introduce a new local variable, `raw_data`, which gets assigned to `args.data` or, if `--file` is used, assigned to the contents of the file. Then `-S` (and all the code below that operates on `args.data`) can be changed to operate on `raw_data`, whether it came from a file or not.

gwbischof:
Updated. Should I add a test?

danielballan:
Looks good. One more question comes to mind as I read this through. In the normal path, it looks like `type(raw_data)` is `str`. In the file-based path, it looks like `type(raw_data)` is `bytes` because the file is read in `'rb'` mode. Offhand, if the goal is push a large array over the wire, a binary option feels right, but as written I think it will blow up:

```py
>>> ast.literal_eval('1')
1
>>> ast.literal_eval(b'1')
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.6/ast.py", line 85, in literal_eval
    return _convert(node_or_string)
  File "/usr/lib/python3.6/ast.py", line 84, in _convert
    raise ValueError('malformed node or string: ' + repr(node))
ValueError: malformed node or string: b'1'
```

danielballan:
One option, of course, is to read as text (`'rb'` -> `'r'`) and think about a binary option separately.

PR #501, opened 2019-08-12
Update pv suffixes

awalter-bnl:
This updates the PV suffixes for setpoint and readback. This is related to a recent change in the IOC that runs the hardware we use this to simulate, and the associated failure of a test we run with this simulator agianst the ophyd object.

The changes to the corresponding ophyd object can be found in `NSLS-II/nslsii` PR#71 [here](https://github.com/NSLS-II/nslsii/pull/71/files)


awalter-bnl:
closing this PR and moving the required test IOC to nsls-ii where it belongs !-)

klauer:
I'm not sure there's harm in keeping the nslsii version in sync (or not). I'd lean slightly toward keeping them the same, but am happy to defer to the decision you made @awalter-bnl.

awalter-bnl:
@klauer I moved the test-ioc to nsls-ii at the suggestion of @danielballan. His point was we should not have to make a change to caproto in order to get a test passing in ophyd due to a change. I agree with this, and hence have just moved a version of this ioc to nslsii.

I am not against removing this entirely from caproto, as it is related to some very specific hardware at NSLS-II I am not sure that caproto is the right place for this anyway.

PR #502, opened 2019-08-12
Address bit rot in master branch.

danielballan:
As noted on the Nikea Slack #caproto channel, I have set up a Travis-CI cron to run `master` daily so that we notice this faster in future.


klauer:
Failing on my patched version of `sendmsg` in trio. Should check to see if `socket.sendmsg` exists upstream yet, and if not, copy over the equivalent to make it work.

Failing that, pinning/removing trio are other options...

danielballan:
I fixed it with the lightest possible touch in 52ba37d. Agreed we should try to swing to `sendmsg`, which if I understand correctly _is_ supported now, but I'll leave that to another PR.

danielballan:
We have at least _one_ passing build (and some still in progress) so all _certain_ failures have been addressed. I move to merge this PR and continue working on the pass rate in follow-up PRs.

danielballan:
P.S. Latest commit ensures we have latest trio. Given that trio is optional/experimental, requiring the latest one feels OK to me.

danielballan:
This seems like it is not a coincidence:

image

All good on R7, failing on 3.14.

danielballan:
No difference. Exactly the same pattern, though, so at least it's reproducible. (Notice that Travis happens to list the builds in a different order, but again the R7 ones passed and the 3-14 ones failed.)

image

danielballan:
All green on the first try! This is 6aa1c30.

image

danielballan:
xref https://github.com/klauer/epics-on-travis/pull/10

danielballan:
Removed `EPICS_MAX_ARRAY_BYTES` `sed` hacks, updated `.ci` commit.

PR #506, opened 2019-08-20
Add support for caproto-put -S.

danielballan:
Previously there was no way to do something like:

```
caproto-put rpi:color 000000
```

where ``000000`` is a string. Now, we can do

```
caproto-put -S rpi:color 000000
```

and it works as intended.

### DIFF
--- a/caproto/commandline/put.py
+++ b/caproto/commandline/put.py
@@ -68,6 +68,8 @@ def main():
                               "space"))
     parser.add_argument('--array-pad', type=int, default=0,
                         help=("Pad the array up to a specified length"))
+    parser.add_argument('-S', action='store_true',
+                        help=("Interpret `data` as a string of characters."))
     parser.add_argument('--no-color', action='store_true',
                         help="Suppress ANSI color codes in log messages.")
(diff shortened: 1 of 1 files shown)


caproto/commandline/put.py
```diff
@@ -68,6 +68,8 @@ def main():
                               "space"))
     parser.add_argument('--array-pad', type=int, default=0,
                         help=("Pad the array up to a specified length"))
+    parser.add_argument('-S', action='store_true',
```
klauer:
```suggestion
    parser.add_argument('-S', dest='as_string', action='store_true',
```

caproto/commandline/put.py
```diff
@@ -84,8 +86,10 @@ def main():
         else:
             set_handler(color=not args.no_color, level='DEBUG')
     logger = logging.LoggerAdapter(logging.getLogger('caproto.ch'), {'pv': args.pv_name})
-
-    if args.array:
+    if args.S:
```
tacaswell:
```suggestion
    if args.as_string:
```

To go with @klauer 's suggestion above.

PR #512, opened 2019-09-17
Allow setting a global timeout via CAPROTO_TIMEOUT env var

mrakitin:
This is based on the discussion in the [Nikea Slack](https://nikea.slack.com/archives/C7QGYD0KW/p1568722811013200). The proposed update allows setting a global timeout via an environment variable `CAPROTO_TIMEOUT`.

### DIFF
--- a/caproto/threading/client.py
+++ b/caproto/threading/client.py
@@ -910,8 +910,9 @@ class Context:
         If None is specified, a fresh one is instantiated.
     timeout : number or None, optional
         Number of seconds before a CaprotoTimeoutError is raised. This default
-        can be overridden at the PV level or for any given operation. If
-        unset, the default is 2 seconds. If None, never timeout.
+        can be overridden at the PV level or for any given operation. If unset,
+        the default is 2 seconds. If None, never timeout. A global timeout can
+        be specified via an environment variable ``CAPROTO_TIMEOUT``.
     host_name : string, optional
         uses value of ``socket.gethostname()`` by default
     client_name : string, optional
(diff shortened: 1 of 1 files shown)


mrakitin:
I'd love to have a test, but not sure how to approach it. https://github.com/caproto/caproto/blob/e1f945a059d189a0231b85fe91e277b48c4bc8d4/caproto/tests/test_cli_scripts.py#L77 comment says to skip the long-running tests. What are they?

danielballan:
I think it's OK to merge this un-tested (once we agree on precedence) because it is easy to understand.

> What are they?

I actually don't know. That comment might be out of date.

mrakitin:
> This gives the environment variable priority over the `timeout` keyword argument. If I instantiate a `Context` with an explicit timeout like `Context(timeout=10)` I might be surprised to find my user input silently ignored.
> 
> What if we give the environment variable lowest precedence? The effect, I believe, would be the same because in practice we rarely (maybe never) actually use the `Context`'s `timeout` parameter, but we would avoid laying a "user input ignored" trap.
> 
> If there is an argument for giving the environment a priority higher than lowest priority, should it be higher still?

There are no any arguments from my side, but I wasn't sure how to workaround those:
- If unset, the default is 2 seconds.
- If None, never timeout.

so I ended up in having the env var a higher priority. What would be the `if` condition given the 2 boundary conditions above? How should it fall back to the env var's value?

danielballan:
Would this work?

```py
GLOBAL_DEFAULT_TIMEOUT = os.environ.get("CAPROTO_TIMEOUT", 2)
# up next to CONTEXT_DEFAULT_TIMEOUT and PV_DEFAULT_TIMEOUT sentinels

...

class Context:
    def __iniit__(..., timeout=GLOBAL_DEFAULT_TIMEOUT, ...):
```

danielballan:
It might be worth including `DEFAULT` in the env var name, as in `CAPROTO_DEFAULT_TIMEOUT`. Just for consideration.

mrakitin:
I like the ideas, thanks @danielballan. Will look at the implementation later today.

mrakitin:
Thanks :)

PR #522, opened 2019-10-06
Release notes for v0.4.1

danielballan:
Let's get a release out to make ``async_lib.Event`` available so that the new 
``worker_thread_pc`` IOC example works on a released version of caproto.


danielballan:
Two builds had one random failure each, not the same one.

Issue #525, opened 2019-10-08
caproto server interface with pyepics


vstadnytskyi:
I have created [my repository](https://github.com/vstadnytskyi/caproto-sandbox/issues), where I explore caproto. I am having an issue with caproto server interfacing pyepics, and I don't understand what is going on. 

Since I am not sure if it is related to caproto or pyepics, I have created an [issue](https://github.com/vstadnytskyi/caproto-sandbox/issues/3) in my repository. 

would you prefer me to copy the entire text here, or is it fine to post a link here?

danielballan:
The link is fine, thanks!

danielballan:
Fixed by #529, as noted in the linked issue.

PR #529, opened 2019-10-09
FIX: add a lock to sending over a VirtualCircuit

tacaswell:
This fixes a bug when:
 - there is a ReadResponse that `sock_sendall` can not send in on go
 - there is a monitor that will fire during that readout

Because the buffer for the ReadResponse is sent in parts, the bytes
for the AddEvent can get sent while the client still thinks it is
receiving the payload from the ReadResponse.  The tail of the payload
will then be read as the next message from the server.

 a) corrupt the read payload
 b) result is a bogus message

---

A minimal example to reproduce the problem is

server:

```python
#!/usr/bin/env python3

import time
from caproto.server import pvproperty, PVGroup, ioc_arg_parser, run
import numpy as np

image_shape = (3960, 3960)


class IOInterruptIOC(PVGroup):
    t1 = pvproperty(value=2.0)
    dt1 = pvproperty(value=0.9, dtype=float, precision=3)
    image = pvproperty(
        value=np.random.randint(0, 256, image_shape).flatten(), dtype=float
    )

    @t1.startup
    async def t1(self, instance, async_lib):
        # Loop and grab items from the queue one at a time
        while True:
            await self.t1.write(time.monotonic())
            await async_lib.library.sleep(0.1)


if __name__ == "__main__":
        default_prefix="big_image:",
        desc="Run an IOC that updates via I/O interrupt on key-press events.",
    )

    ioc = IOInterruptIOC(**ioc_options)
    print(ioc.image)
    run(ioc.pvdb, **run_options)

```

client:
```python
#!/usr/bin/env python3
# -*- coding: utf-8 -*-
"""

"""
import epics

# import caproto.threading.pyepics_compat as epics

default_prefix = "big_image:"

image = epics.PV(pvname=default_prefix + "image")
t1 = epics.PV(pvname=default_prefix + "t1")
t1.add_callback(lambda value, **kwargs: print(f"t1: {value}"))

for j in range(42):
    img = image.get(timeout=20)
    print(img.shape)

```

Not sure we want to put in a CI test for this as in is a _bit_ flaky and definitely slow.

----

attn @vstadnytskyi

### DIFF
--- a/caproto/asyncio/server.py
+++ b/caproto/asyncio/server.py
@@ -82,7 +82,7 @@ class VirtualCircuit(_VirtualCircuit):
 
             async def recv(self, nbytes):
                 return (await self.loop.sock_recv(self.client, 4096))
-
+        self._raw_lock = asyncio.Lock()
         self._raw_client = client
         super().__init__(circuit, SockWrapper(loop, client), context)
         self.QueueFull = asyncio.QueueFull
@@ -109,8 +109,11 @@ class VirtualCircuit(_VirtualCircuit):
     async def send(self, *commands):
         if self.connected:
             buffers_to_send = self.circuit.send(*commands)
-            await self.loop.sock_sendall(self._raw_client,
-                                         b''.join(buffers_to_send))
+            # lock to make sure a AddEvent does not write bytes
+            # to the socket while we are sending
+            async with self._raw_lock:
+                await self.loop.sock_sendall(self._raw_client,
+                                             b''.join(buffers_to_send))
 
     async def run(self):
         self._cq_task = self.loop.create_task(self.command_queue_loop())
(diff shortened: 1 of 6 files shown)


danielballan:
Nice quick fix! Before merging we should verify that `acctst` still runs.

danielballan:
Could this affect the other servers? It looks like trio has a lock for this (on phone, can’t easily link...) but not sure.

klauer:
Yes, all will need modification.

tacaswell:
Fixed everywhere.

tacaswell:
closes #525

tacaswell:
I suspect I don't understand something about how timeouts propagate...

danielballan:
The timeout propagation fix is now in #533. I'll merge that first, once CI passes, and then rebase this.

danielballan:
One random failure:

```
caproto/tests/test_pyepics_compat.py::test_emptyish_char_waveform_no_monitor FAILED 
```

Re-running to be sure...

But `acctst` (run locally, as usual) has passed.

danielballan:
Failure as not repeated on re-run. Merging!

PR #531, opened 2019-10-09
TST: be more forgiving while trying to start the test server

tacaswell:
We seem to hit the trio servers before they are ready.


danielballan:
Good find! It did not occur to me that we could get errors other than `TimeoutError` in this case.

Issue #537, opened 2019-10-25
mock records not compliant to 'real' EPICS records


baldwinb:
I have found that reading all of the fields of a mock record show the following problems:
1) a read of a string typically returns a char array/waveform (when linux cli expects string)
2) some PVs fail read of value, or read of enums

For more information:

1) 
A change is needed to allow:
- return of string PVs as "DBR_STRING" type when caget from cli without appending "$"
- return of char array "DBR_CHAR" type when caget from cli when appending "$"

Appending "$" to the PV should return a char array. Typically this is used from linux CLI: 
caget -S "PV$"
so that the char array is converted to a string, and this has no length requirements. Caproto currently returns char array for both.

Refer to pull request from https://github.com/baldwinb/caproto/tree/compliance_check , which adds compliance test highlighting other issues that can be observed when reading all pvs of a mock record from linux cli, or common epics gui tools. The pull request also changes the records.py file to use string type instead of char array (which is how EPICs normally works). However, modifying records.py to use strings means it now returns strings even when appending "$".

2)
Other miscellaneous problems are also evident when running "pytest test_ai_compliance.py" that need to be fixed (but is not clear to me how to do this).  Such as: problems reading ADEL field, and enum of STAT field.

klauer:
Fixing the ADEL issue with: https://github.com/klauer/caproto/commit/4ea0d6cb1459b9cf5a7b9485bcd53e19d9487d55

klauer:
I think we're largely there with regards to compliance. Thanks for the push to get this going @baldwinb.

Let's open up new issues for specific records/fields as we find them.

PR #538, opened 2019-10-25
improve mock 'ai' record compliance to 'real' EPICS record 

baldwinb:
Changed records.py so that common pv strings favors string type instead of char type. (Refer to mock record issue).

Added a test function that fetches real 'ai' record fields from the web and types, and then:
-compares types between mock ai and real ai
-ensures that all the mock fields can be read
-ensures any mock record enums can be read
-reads "PV$" as a string, and "PV" (without "$") as native type, (reflecting how EPICS would normally perform these reads)

You can test separately by "pytest test_ai_compliance.py", showing differences/problems with the mock record.

See related issue: https://github.com/caproto/caproto/issues/537


tacaswell:
Thanks for this!  

Do we have to use pyepics in the tests or can we use the caproto threading client?

The test failures look like conflicts between {pytest, trio, attrs}.

baldwinb:
I think that is fine if it gives the same errors.

The python test I wrote highlights the same problems that I observed when trying to read the pvs using our real epics ui; "pv properties" ( https://qtepics.github.io/ ).

"pv properties" polls all the .* PVs of a record for display in a list using real epics. When the ui is opened it showed:
-errors generated in the caproto server stdout/err (I originally ran in debug, via windows with asyncio)
-in particular with STAT enum, read issues on MDEL and ADEL
-RTYP is returned as char array, which the gui uses to decide which other .* PVs to look for. ie returns [97, 105] instead of 'ai'. ( the ui also tries to read RTYP$ which did not work at the time)

 If you run the test function without pytest you can see it is not a pytest issue. You could also try on a "real" record (which I did and had one short <> long error for ROFF, which is typically not used).

I am not sure there is a consequence of difference in type between short and long (probably not for us) but I have left it in as a fail because it is technically not compliant.

klauer:
Merged as part of #577 - thanks @baldwinb!

danielballan:
:tada:

Issue #540, opened 2019-10-30
sync and threading clients raise error on Windows 7


vstadnytskyi:
I just get an error when I am trying to use Threading client on Windows 7 machine. I do have the latest caproto available which is 0.4.1

```
pip3 install --upgrade caproto
Requirement already up-to-date: caproto in c:\users\user\appdata\local\prog
rams\python\python37\lib\site-packages (0.4.1)
```

My python code:
```
Python 3.7.4 (tags/v3.7.4:e09359112e, Jul  8 2019, 20:34:20) [MSC v.1916 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license()" for more information.
>>> import caproto
>>> caproto.__version__
'0.4.1'
>>> from caproto.threading.client import Context
>>> ctx = Context()
Traceback (most recent call last):
  File "<pyshell#3>", line 1, in <module>
    ctx = Context()
  File "C:\Users\femtoland\AppData\Local\Programs\Python\Python37\lib\site-packages\caproto\threading\client.py", line 936, in __init__
    broadcaster = SharedBroadcaster()
  File "C:\Users\femtoland\AppData\Local\Programs\Python\Python37\lib\site-packages\caproto\threading\client.py", line 342, in __init__
    self.broadcaster.our_address = self.udp_sock.getsockname()[:2]
OSError: [WinError 10022] An invalid argument was supplied
>>> 
```

vstadnytskyi:
There is something funny with Windows 7. The sync client doesn't work either. But It works on mac.

Windows:
```
Python 3.7.4 (tags/v3.7.4:e09359112e, Jul  8 2019, 20:34:20) [MSC v.1916 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license()" for more information.
>>> import caproto
>>> caproto.__version__
'0.4.1'
>>> from caproto.sync.client import read
>>> read('NIH:TEMP.RBV')
Traceback (most recent call last):
  File "<pyshell#5>", line 1, in <module>
    read('NIH:TEMP.RBV')
  File "C:\Users\femtoland\AppData\Local\Programs\Python\Python37\lib\site-packages\caproto\sync\client.py", line 256, in read
    chan = make_channel(pv_name, udp_sock, priority, timeout)
  File "C:\Users\femtoland\AppData\Local\Programs\Python\Python37\lib\site-packages\caproto\sync\client.py", line 133, in make_channel
    address = search(pv_name, udp_sock, timeout)
[...]

Issue #514, opened 2019-09-19
Resolve OSError from premature call to getsockname() on Windows


danielballan:
Starting in v0.4.0, we call `sock.getsockname()` in many places to obtain "our address" for logging purposes. Apparently, we sometimes do this on an _unbound_ socket, which [reportedly raises an error "on Windows"](https://mail.python.org/pipermail/python-list/2015-January/683777.html) but I guess returns without an error on Linux/OSX. (Not sure what it would return....)

This error on Windows was observed and reported in by @mrakitin and @EliotGann in https://github.com/nsls-ii-forge/caproto-feedstock/issues/2. The key section is:

```
~\.conda\envs\collection-nsls2forge-py37-defaults\lib\site-packages\caproto\threading\pyepics_compat.py in PV()
    170                'upper_warning_limit', 'upper_ctrl_limit', 'lower_ctrl_limit',
    171                'put_complete')
--> 172     _default_context = Context(SharedBroadcaster())
    173
    174     def __init__(self, pvname, callback=None, form='time',

~\.conda\envs\collection-nsls2forge-py37-defaults\lib\site-packages\caproto\threading\client.py in __init__(self, registration_retry_time)
    340
    341         self.broadcaster = ca.Broadcaster(our_role=ca.CLIENT)
--> 342         self.broadcaster.our_address = self.udp_sock.getsockname()[:2]
    343         self.log = logging.LoggerAdapter(
    344             self.broadcaster.log, {'role': 'CLIENT'})

OSError: [WinError 10022] An invalid argument was supplied

In [1]:
```

Our intent was to only call `getsockname()` after the socket is bound. There a couple places where the code goes out of its way to call it as early as possible but no earlier. It looks like we didn't get that quite right. The best fix would be to avoid calling `getsockname()` on an unbound socket. If we can't figure out that, a stop-gap could be to make some `safe_getsockname` that swallows the `OSError`.

cc @ke-zhang-rd

mrakitin:
Thanks for documenting it @danielballan!

PR #543, opened 2019-11-08
MNT: Bind to PORT 0

ke-zhang-rd:
**Goal**
This PR is trying to solve #540 and #514 related socket error only reported on windows. 

**Solution**
It looks like Linux and Mac will assign `('0.0.0.0', 0)` to socket, if we didn't explicit `bind`.

`bind` argument in this PR is based on 
- https://eklitzke.org/binding-on-port-zero (I also have  same concern mentioned in this article)
- https://docs.python.org/3/library/socket.html#socket.create_connection

**Test**
This PR solved scenario in those two issues(#540, #514) but may need further tests.

### DIFF
--- a/caproto/threading/client.py
+++ b/caproto/threading/client.py
@@ -340,8 +340,7 @@ class SharedBroadcaster:
         self.listeners = weakref.WeakSet()
 
         self.broadcaster = ca.Broadcaster(our_role=ca.CLIENT)
-        self.udp_sock.bind(('', 0))
-        self.broadcaster.our_address = self.udp_sock.getsockname()[:2]
+        self.broadcaster.client_address = None
         self.log = logging.LoggerAdapter(
             self.broadcaster.log, {'role': 'CLIENT'})
         self.search_log = logging.LoggerAdapter(
@@ -443,7 +442,6 @@ class SharedBroadcaster:
         """
         bytes_to_send = self.broadcaster.send(*commands)
         tags = {'role': 'CLIENT',
-                'our_address': self.broadcaster.our_address,
                 'direction': '--->>>'}
         for host in ca.get_address_list():
             if ':' in host:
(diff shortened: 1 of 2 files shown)


danielballan:
The concern mentioned in the article is that if you bind to the port, close the socket and then try binding to the same port again (perhaps in a separate “program”) the port may not still be available. Does that apply here?

danielballan:
Actually I think there is a deeper problem here. We use this socket to broadcast to multiple interfaces

https://github.com/caproto/caproto/blob/a3fb3d5cc6ae2f8817904b801e70497283509cb5/caproto/threading/client.py#L459

so there is no single “our address”. I have an unfinished branch I worked on at ICALEPCS that might be helpful. Will push when I’m back to a laptop.

ke-zhang-rd:
> The concern mentioned in the article is that if you bind to the port, close the socket and then try binding to the same port again (perhaps in a separate “program”) the port may not still be available. Does that apply here?

I don't think so and want to double check with you.

danielballan:
Gotcha. I don't think so either but wanted to double-check with you! :-D

danielballan:
I think #544 and especially the change from `our_address` and `their_address` to `our_addresses` and `their_addresses` at the top of this commit https://github.com/caproto/caproto/pull/544/commits/ff8a91ae7a2c2df01d6a85ba02d6e8b0e5344846 will help.

danielballan:
I don't _think_ it fixes the issue, but it's the right framework in which to fix the issue. Can you test and see if the bug is still there, to be sure?

ke-zhang-rd:
I was trying to say bind socket with `''`(empty string) might be right based figure I found here
https://realpython.com/python-sockets/#using-hostnames. Then we use `getsockname` to access address when we need it and we don't need to init `our_address` in `__init__`
image
 
I might also totally wrong.
I will check #544

caproto/threading/client.py
```diff
@@ -670,7 +672,7 @@ def command_loop(self):
                                             name, cid, *accepted_address, *new_address,
                                             extra={'pv': name,
                                                    'their_address': accepted_address,
-                                                   'our_address': self.udp_sock.getsockname()[:2]})
+                                                   'our_address': self.broadcaster.client_address})
```
danielballan: This won't work until #544 is merged, right? I don't think `client_address` exists yet.

```
$ git grep client_address
caproto/_commands.py:    .. attribute:: client_address
caproto/_commands.py:    def __init__(self, client_address='0.0.0.0'):
caproto/_commands.py:            ipv4_to_int32(str(client_address)))
caproto/_commands.py:    def client_address(self):
caproto/tests/test_serialization.py:    'client_address': [ip],
doc/source/command-line-client.rst:    [D 20:00:47.564 _broadcaster:74] 1 of 1 RepeaterRegisterRequest(client_address='0.0.0.0')
doc/source/shark.rst:   1550679067.619182 192.168.86.21:55928->255.255.255.255:5065 RepeaterRegisterRequest(client_address='0.0.0.0')
```

ke-zhang-rd: I mean using `our_address` here. #544 isn't required.

danielballan:
Revisiting this, I think your first approach was the right one. Refining that a bit, if we bind right after we create the socket we can avoid getting tangled up in the `our_addresses` vs `client_address` code that follows. We can be sure that all calls to `getsockname()` from then on will work. See #551.

klauer:
Is there still work to do here now that #551 is in?

danielballan:
I think not. Please reopen if I am mistaken, @ke-zhang-rd.

Issue #545, opened 2019-11-15
ReadNotify and EventAdd Responses do not contain PV information


vstadnytskyi:
I have made one callback function that I call whenever a subscribed value changes its' value on a server. The idea is that I can just pass that information to a code that will use it and act on it if needed. In this way I can write one module that will act as input/output(IO) and simply pass the incoming information to the main system code(system. If something needs to be published the system sends information to IO and it takes care of it.  However, I discovered that ReadNotifyResponse a resulting object from read() and EventAddResponse a resulting object from subscribe() do not contain information about the PV. Hence, I cannot distinguish between callbacks generated by different PVs. 

Is there a reason why Response objects don't have PV information?

Examples:

If I subscribe to a PV and have a callback function that just prints the content of the response 
"self.sub.add_callback(self.monitor)"
```
def monitor(self,response):
        print(f'system monitor received {response}')

system monitor received EventAddResponse(data=array([2.37]), data_type=<ChannelType.DOUBLE: 6>, data_count=1, status=CAStatusCode(name='E
CA_NORMAL', code=0, code_with_severity=1, severity=<CASeverity.SUCCESS: 1>, success=1, defunct=False, description='Normal successful comp
letion'), subscriptionid=6, metadata=None)
```
and 
if I try to read the PV with sync client  "sync.client.read()"
```
pv.read()
Out[12]: ReadNotifyResponse(data=array([2.37]), data_type=<ChannelType.DOUBLE: 6>, data_count=1, status=CAStatusCode(name='ECA_NORMAL', c
ode=0, code_with_severity=1, severity=<CASeverity.SUCCESS: 1>, success=1, defunct=False, description='Normal successful completion'), ioi
d=7, metadata=None)

```

vstadnytskyi:
Something else I found related to ReadNotifyResponse.
Somehow there are attributes with None values. Is it a bug or intentional and reserver for future expansion?
 
```
In [35]: resp = client.pvs['TOWER.P1.BUSY'].read()
In [36]: resp
Out[36]: ReadNotifyResponse(data=array([0], dtype=uint16), data_type=<ChannelType.ENUM: 3>, data_count=1, status=CAStatusCode(name='ECA
_NORMAL', code=0, code_with_severity=1, severity=<CASeverity.SUCCESS: 1>, success=1, defunct=False, description='Normal successful comp
letion'), ioid=11, metadata=None)

In [37]: type(resp.sender_address)
Out[37]: NoneType
```

[...]

PR #547, opened 2020-01-16
Provide Subscription to user callback.

danielballan:
In #545, @vstadnytskyi pointed out that it is inconvenient that subscription
updates do not carry information about which PV they pertain to. @tacaswell
suggested that we could change the signature of user callbacks to include the
PV as well as the raw response. Getting even more specific, we can include
the Subscription object `sub`, from which you can access the PV as `sub.pv` and the PV
name as `sub.pv.name`. That is, the callback signature would change from
``f(response)`` to ``f(response, sub)``.

(An alternative proposal, adding the PV name to the response, was rejected
because we have held that Message objects contain nothing more or less
than the information sent over the wire.)

Fortunately, this change can be made in a backward-compatible way. When a
user callback is first registered

```py
sub.add_callback(f)
```

we can inspect the signature of ``f`` *once* to determine if it would accept
only one argument. If so, we wrap ``f`` in a function that drops
the new ``sub`` argument and passes only ``response`` through to ``f`` and
subscribe that wrapper function instead. We also issue a warning.

This has not been tested locally, may need more work.

### DIFF
--- a/caproto/threading/client.py
+++ b/caproto/threading/client.py
@@ -2167,7 +2167,7 @@ class Subscription(CallbackHandler):
         # As implemented below, updates are blocking further messages from
         # the CA servers from processing. (-> ThreadPool, etc.)
         pv = self.pv
-        super().process(command, self)
+        super().process(self, command)
         self.log.debug("%r: %r", pv.name, command)
         self.most_recent_response = command
 
@@ -2178,7 +2178,7 @@ class Subscription(CallbackHandler):
         Parameters
         ----------
         func : callable
-            Expected signature: ``func(response, sub)``.
+            Expected signature: ``func(sub, response)``.
             The signature ``func(response)`` is also supported for
             backward-compatibility but will issue warnings. Support will be
             removed in a future release of caproto.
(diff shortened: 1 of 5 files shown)


danielballan:
Closes #545

klauer:
Love the signature check back-compat. 👍

caproto/threading/client.py
```diff
@@ -2160,13 +2164,51 @@ def add_callback(self, func):
...
         -------
         token : int
             Integer token that can be passed to :meth:`remove_callback`.
+
+        .. versionchanged:: 0.5.0
+
+           Changed the expected signature of ``func`` to add ``pv``.
         """
+        # Handle func with signature func(respons) for back-compat.
+        if len(inspect.signature(func).parameters) == 1:
```
klauer: 2 tests are failing due to this:

`def callback(command, **kwargs):`
having a signature:
```
In [2]: sig.parameters
Out[2]:
mappingproxy({'command': <Parameter "command">,
              'kwargs': <Parameter "**kwargs">})
```

Might consider using `sig.bind(None)` to see if it takes only one positional arg, or `sig.bind(None, None)` (for example) to see if it has the new signature.

danielballan: (The `f(command, **kwargs)` example lives on a new, second test of the deprecated code path.)

PR #552, opened 2020-01-27
Direct fix for #514

### LINKED ISSUE #514: Resolve OSError from premature call to getsockname() on Windows

danielballan:
Our discussion around how to fix #514 has led to some clarity but also some open questions. I think it's fair to say that no one involved in the discussion so far is feeling confident that we understand the implications deeply enough to prescribe the single correct way forward. Rather than picking one, I suggest we apply this targeted workaround to the specific issue that broke caproto on Windows in v0.4.0. The discussion can continue, perhaps pulling in new expertise, and we can go from there, hopefully eventually retiring this workaround.

If there are no strong objections, I would like to merge this and release a working 0.4.x release this week.

### DIFF
--- a/caproto/curio/client.py
+++ b/caproto/curio/client.py
@@ -12,6 +12,7 @@ import getpass
 import logging
 
 import caproto as ca
+from caproto._utils import safe_getsockname
 import curio
 
 from collections import OrderedDict
@@ -237,7 +238,7 @@ class SharedBroadcaster:
 
         # UDP socket broadcasting to CA servers
         self.udp_sock = ca.bcast_socket(socket)
-        self.broadcaster.our_address = self.udp_sock.getsockname()[:2]
+        self.broadcaster.our_address = safe_getsockname(self.udp_sock)[:2]
         self.registered = False  # refers to RepeaterRegisterRequest
         self.loop_ready_event = curio.Event()
         self.unanswered_searches = {}  # map search id (cid) to name
--- a/caproto/curio/client.py
+++ b/caproto/curio/client.py
@@ -41,7 +41,7 @@ class VirtualCircuit:
     async def connect(self):
         async with self._socket_lock:
             self.socket = await socket.create_connection(self.circuit.address)
-            self.circuit.our_address = self.socket.getsockname()[:2]
+            self.circuit.our_address = self.socket.getsockname()
             # Kick off background loops that read from the socket
             # and process the commands read from it.
             await curio.spawn(self._receive_loop, daemon=True)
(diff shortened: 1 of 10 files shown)


caproto/curio/client.py
```diff
@@ -237,7 +238,7 @@ def __init__(self):
 
         # UDP socket broadcasting to CA servers
         self.udp_sock = ca.bcast_socket(socket)
-        self.broadcaster.our_address = self.udp_sock.getsockname()[:2]
+        self.broadcaster.our_address = safe_getsockname(self.udp_sock)[:2]
```
danielballan: @ke-zhang-rd Do you recall why we have `[:2]` everywhere this is used? Doesn't this _always_ return a `(host, port)` tuple?

ke-zhang-rd: I remembered I get this from some online examples. However, it's necessary. Check [here].(https://github.com/python/cpython/blob/master/Lib/socket.py#L633-L634)

danielballan: Thanks for the reference, very helpful. CA is not compatible with IPv6; if we encounter IPv6, other, more essential aspects of caproto will break. I should have asked for clarification on this in the initial review. These should all be removed.

ke-zhang-rd: I'm ok.

danielballan: Alright. All removed in 25cbb62. Confirmed:

```
$ git grep -e "\[:2\]"
caproto/_commands.py:        return DBR_INT.from_buffer(bytearray(self.buffers[0])[:2]).value
```

danielballan:
Approved by @tacaswell and @klauer over private channels. Self-merging to move this along so we can finally get a fix out.

PR #557, opened 2020-02-03
FIX: validate enum strings

tacaswell:
There are limits to both the length of the enum strings and the number
of enum
entries (https://github.com/epics-base/epics-base/blob/d0c4cc0cec8d31b88ee480d2d7f007f17cf0cdbf/modules/ca/src/client/db_access.h#L38-L40)
which we do have encoded in `_constants.py` and respect on sending the
value out, but we should reject invalid values being sent on the way in.

### DIFF
--- a/caproto/_data.py
+++ b/caproto/_data.py
@@ -714,10 +714,10 @@ class ChannelEnum(ChannelData):
     @staticmethod
     def _validate_enum_strings(enum_strings):
         if any(len(es) >= MAX_ENUM_STRING_SIZE for es in enum_strings):
-            over_lenth = tuple(f'{es}: {len(es)}' for es in enum_strings if
-                               len(es) >= MAX_ENUM_STRING_SIZE)
+            over_length = tuple(f'{es}: {len(es)}' for es in enum_strings if
+                                len(es) >= MAX_ENUM_STRING_SIZE)
             msg = (f"The maximum enum string length is {MAX_ENUM_STRING_SIZE} " +
-                   f"but the strings {over_lenth} are too long")
+                   f"but the strings {over_length} are too long")
             raise ValueError(msg)
         if len(enum_strings) > MAX_ENUM_STATES:
             raise ValueError(f"The maximum number of enum states is {MAX_ENUM_STATES} " +
--- a/caproto/_data.py
+++ b/caproto/_data.py
@@ -17,7 +17,7 @@ from ._dbr import (DBR_TYPES, ChannelType, native_type, native_types,
 from ._utils import CaprotoError, CaprotoValueError, ConversionDirection
 from ._commands import parse_metadata
 from ._backend import backend
-
+from ._constants import MAX_ENUM_STRING_SIZE, MAX_ENUM_STATES
 
 __all__ = ('Forbidden',
            'ChannelAlarm',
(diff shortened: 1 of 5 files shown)


tacaswell:
This is failing because the enums generated in the mock-records have more than 16 entries.

Trying to get the emunstrings fails with the error 

```python
Server failed to process command: ReadNotifyRequest(data_type=<ChannelType.CTRL_ENUM: 31>, data_count=0, sid=1, ioid=6)
  File "/home/tcaswell/src/bnl/epics/caproto/caproto/server/common.py", line 188, in _command_queue_iteration
    response = await self._process_command(command)
  File "/home/tcaswell/src/bnl/epics/caproto/caproto/server/common.py", line 449, in _process_command
    data_type, user_address=self.circuit.address)
  File "/home/tcaswell/src/bnl/epics/caproto/caproto/_data.py", line 385, in auth_read
    return (await self.read(data_type))
  File "/home/tcaswell/src/bnl/epics/caproto/caproto/server/server.py", line 127, in read
    return await self._read(data_type)
  File "/home/tcaswell/src/bnl/epics/caproto/caproto/_data.py", line 417, in _read
    self._read_metadata(dbr_metadata)
  File "/home/tcaswell/src/bnl/epics/caproto/caproto/_data.py", line 752, in _read_metadata
    for s in self.enum_strings]
  File "/home/tcaswell/src/bnl/epics/caproto/caproto/_dbr.py", line 648, in enum_strings
    self.strs[i][:] = bytes_.ljust(MAX_ENUM_STRING_SIZE, b'\x00')
IndexError: invalid index
```
generated via

```python
# python caproto/ioc_examples/scalars_and_arrays.py --list-pvs
pvm = caproto.threading.pyepics_compat.get_pv('arr:enum.NSTA')
pvm.enum_strs
```

which is because we per-allocate only 16 slots for enum strings but then try to shove in more than 16 strings.

@klauer need some reference on what rsrv does in this case.

danielballan:
Possibly useful: https://epics.anl.gov/tech-talk/2012/msg01574.php

caproto/_data.py
```diff
@@ -711,12 +711,25 @@ def check_access(self, hostname, username):
 class ChannelEnum(ChannelData):
     data_type = ChannelType.ENUM
 
+    @staticmethod
+    def _validate_enum_strings(enum_strings):
+        if any(len(es) >= MAX_ENUM_STRING_SIZE for es in enum_strings):
+            over_lenth = tuple(f'{es}: {len(es)}' for es in enum_strings if
```
klauer:
```suggestion
            over_length = tuple(f'{es}: {len(es)}' for es in enum_strings if
```

klauer:
It's not really possible in a normal IOC to have more than our stated maximum. I don't think documentation or further investigation is particularly warranted.

tacaswell:
I fixed the spelling, but the tests are stil failing due to https://github.com/caproto/caproto/blob/1a7356487fd930cb6dee6c26f1ce1f117cd87c95/caproto/server/menus.py#L237-L259 sourceing too many enum values to the records in a whole bunch of places such as https://github.com/caproto/caproto/blob/1a7356487fd930cb6dee6c26f1ce1f117cd87c95/caproto/server/records.py#L344-L349

What should be done about those enum records?

klauer:
The `.STAT` field isn't a record; it's a field of a record. While that may sound pedantic, it's an important distinction. There are no records which allow for more than that maximum number of enum values.

For fields like `.STAT` which have more than the maximum number of enum values, it appears that CA just truncates the remainder when queried:
```
13SIM1:Stats5:Reset1.STAT
    Native data type: DBF_ENUM
    Request type:     DBR_CTRL_ENUM
    Element count:    1
    Value:            Illegal Value (17)
    Status:           UDF
    Severity:         INVALID
    Enums:            (16)
                      [ 0] NO_ALARM
                      [ 1] READ
                      [ 2] WRITE
                      [ 3] HIHI
                      [ 4] HIGH
                      [ 5] LOLO
                      [ 6] LOW
                      [ 7] STATE
                      [ 8] COS
                      [ 9] COMM
                      [10] TIMEOUT
                      [11] HWLIMIT
                      [12] CALC
                      [13] SCAN
                      [14] LINK
                      [15] SOFT
```

tacaswell:
You know I am always on board for pedantic distinctions :)

It was my understanding was that you could only ever really talk to a field (and a record is a bunch of fields) so I am not sure how to understand a "record" having a enums as we can only talk to fields.

caproto/server/records.py
```diff
@@ -122,7 +123,7 @@ class RecordFieldGroup(PVGroup):
     new_alarm_status = pvproperty(
         name='NSTA',
         dtype=ChannelType.ENUM,
-        enum_strings=menus.menuAlarmStat.get_string_tuple(),
+        enum_strings=menus.menuAlarmStat.get_string_tuple()[:MAX_ENUM_STATES],
```
danielballan: It seems like it could be easy to forget to add this. Should it be baked into `get_string_tuple()`? Or exposed as an option to that method?

klauer: That sounds like a reasonable solution, @danielballan.

klauer: Addressed in 2234905

tacaswell:
Thanks @klauer  I had started to work on it but got way-laid by things as I normally do...

klauer:
No worries, @tacaswell

klauer:
Aside: I mistakenly pushed 623c994 here (meant for the other PR). Oh well, it's likely failing here too.

klauer:
Azure all passing, and I'm impatient to wait for travis. (YOLO numero dos)

PR #565, opened 2020-02-28
Check if data is of type str

### LINKED ISSUE #564: caproto-put fails to assign enum type PVs
Cannot find a way to successfully assign enum PVs using `caproto-put`.

`caproto-get` works as expected.

Easy way to reproduce if `docker` available:

```bash
# Terminal 1
$ docker run -it dmscid/lewis chopper
(Ctrl-C to quit)

# Terminal 2
$ caproto-get AutoPark
AutoPark                                  [false]
$ caproto-get -d int AutoPark
AutoPark                                  [0]
$ caproto-put AutoPark 1
'int' object is not subscriptable
$ caproto-put AutoPark true
invalid literal for int() with base 10: 'true'
$ caproto-get AutoPark
AutoPark                                  [false]
```

To verify PV type and possible values:
```bash
$ caget -d 31 AutoPark
AutoPark
    Native data type: DBF_ENUM
    Request type:     DBR_CTRL_ENUM
    Element count:    1
    Value:            false
    Status:           NO_ALARM
    Severity:         NO_ALARM
    Enums:            ( 2)
                      [ 0] false
                      [ 1] true
```
Using `caget` here because `caproto-get` seems to ignore option `-d 31` / request type `DBR_CTRL_ENUM`.

MikeHart85:
Closes #564.

In Python3, checking if a variable `isinstance` of `str` should be sufficient to determine stringyness.

The previous condition prevented enum type PVs from being set.

### DIFF
--- a/caproto/sync/client.py
+++ b/caproto/sync/client.py
@@ -548,7 +548,9 @@ def _write(chan, data, metadata, timeout, data_type, notify):
         stringy_data = False
         if isinstance(data, (str, bytes)):
             stringy_data = True
-        if hasattr(data, '__getitem__') and isinstance(data[0], (str, bytes)):
+        if hasattr(data, '__getitem__') \
+                and len(data) > 0 \
+                and isinstance(data[0], (str, bytes)):
             stringy_data = True
 
         if stringy_data:
--- a/caproto/sync/client.py
+++ b/caproto/sync/client.py
@@ -542,10 +542,16 @@ def _write(chan, data, metadata, timeout, data_type, notify):
     logger.debug("Detected native data_type %r.", chan.native_data_type)
     # abundance of caution
     ntype = field_types['native'][chan.native_data_type]
-    if data_type is None:
-        # Handle ENUM: If data is INT, carry on. If data is STRING,
-        # write it specifically as STRING data_Type.
-        if (ntype is ChannelType.ENUM) and isinstance(data, str):
+    if (data_type is None) and (ntype is ChannelType.ENUM):
+        # Change data_type to STRING if data contains string-like data, or
+        # iterable of string-like data
+        stringy_data = False
+        if isinstance(data, (str, bytes)):
+            stringy_data = True
+        if hasattr(data, '__getitem__') and isinstance(data[0], (str, bytes)):
+            stringy_data = True
+
+        if stringy_data:
             logger.debug("Will write to ENUM as data_type STRING.")
             data_type = ChannelType.STRING
     logger.debug("Writing.")
@@ -597,7 +603,7 @@ def write(pv_name, data, *, notify=False, data_type=None, metadata=None,
     ----------
     pv_name : str
         The PV name to write to
-    data : str, int, or float or any Iterable of these
+    data : str, bytes, int, or float or any Iterable of these
         Value(s) to write.
     notify : boolean, optional
         Request notification of completion and wait for it. False by default.
@@ -680,7 +686,7 @@ def read_write_read(pv_name, data, *, notify=False,
     ----------
     pv_name : str
         The PV name to write/read/write
-    data : str, int, or float or a list of these
+    data : str, bytes, int, or float or any Iterable of these
         Value to write.
     notify : boolean, optional
         Request notification of completion and wait for it. False by default.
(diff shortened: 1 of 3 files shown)


tacaswell:
knee-jerk comments from the :peanuts: gallery.

I think this is also colliding with the details about "should a scalar be a single value or a length 1 array in the code base".

Internally I thought we kept everything as bytes rather than `str`?  There are some issues with encoding and dealing with when we go to (iirc) latin-1 to push the strings out to the wire.

MikeHart85:
It definitely comes in as a single value in this scenario, but whether it should is a different question.

> Internally I thought we kept everything as bytes rather than str? There are some issues with encoding and dealing with when we go to (iirc) latin-1 to push the strings out to the wire.

Yes, that part struck me as odd as well.

After applying this change, it works, but note the type on the old and new values:

```bash
$ caproto-put AutoPark 1
Old : AutoPark                                  [b'false']
New : AutoPark                                  [b'true']
$ caproto-put AutoPark false
Old : AutoPark                                  [b'true']
New : AutoPark                                  [b'false']
```

Also:

```python
>>> type(b'foobar')
<class 'bytes'>
>>> isinstance(b'foobar', str)
False
```

So, at that particular point in the code, `data` is definitely of type `str` and not `bytes`. But in other parts of the system it is kept as `bytes`, and presented to the user that way.

Should we expand this PR to investigate a bit more and see if we can make that more consistent, or leave that for another time and consider this a small bugfix for this specific scenario? The change as is shouldn't affect anything other than enum PVs, since that's also part of the condition.

MikeHart85:
Looks like I spoke too soon... judging by those test failures, this breaks other usage scenarios. I'll have a closer look.

MikeHart85:
I don't think most (all?) of the test failures from before had anything to do with this change. They look like timeouts waiting for fake IOCs to spin up, which I don't think can be related?

I've expanded type checking such that the condition will pass for any string, byte string, or iterable thereof. This should bring expectations of possible inputs in line with the documentation of `write` and `read_write_read` (which I've also updated to include `bytes`):

```
    data : str, bytes, int, or float or any Iterable of these
        Value(s) to write.
```

The condition was starting to get a bit convoluted, since you have to watch out for `bytes` being difficult to distinguish from a list of ints, so I decided to inwind it and make it a bit more verbose but hopefully also easier to follow.

caproto/sync/client.py
```diff
@@ -542,10 +542,16 @@ def _write(chan, data, metadata, timeout, data_type, notify):
...
     # abundance of caution
     ntype = field_types['native'][chan.native_data_type]
-    if data_type is None:
-        # Handle ENUM: If data is INT, carry on. If data is STRING,
-        # write it specifically as STRING data_Type.
-        if (ntype is ChannelType.ENUM) and isinstance(data[0], bytes):
+    if (data_type is None) and (ntype is ChannelType.ENUM):
+        # Change data_type to STRING if data contains string-like data, or
+        # iterable of string-like data
+        stringy_data = False
```
danielballan: Easier to follow. Also, I love this variable name.

caproto/sync/client.py
```diff
@@ -542,10 +542,16 @@ def _write(chan, data, metadata, timeout, data_type, notify):
...
+        if isinstance(data, (str, bytes)):
+            stringy_data = True
+        if hasattr(data, '__getitem__') and isinstance(data[0], (str, bytes)):
```
klauer: `data = tuple()` I believe would be problematic. Should add condition for `len(data)` as well

MikeHart85: I hadn't thought of that at all, good catch.

MikeHart85:
Test failures still look like timeouts that aren't likely to be related (after merging updated master).

Command line behavior seems to be as expected now:

```bash
$ docker run -itd --rm --name test_ioc dmscid/lewis chopper
$ cd /path/to/caproto
$ python -m caproto.commandline.get AutoPark
AutoPark                                  [false]
$ python -m caproto.commandline.put AutoPark 1
Old : AutoPark                                  [b'false']
New : AutoPark                                  [b'true']
$ python -m caproto.commandline.put AutoPark 0
Old : AutoPark                                  [b'true']
New : AutoPark                                  [b'false']
$ python -m caproto.commandline.put AutoPark true
Old : AutoPark                                  [b'false']
New : AutoPark                                  [b'true']
$ python -m caproto.commandline.put AutoPark false
Old : AutoPark                                  [b'true']
New : AutoPark                                  [b'false']
$ docker stop test_ioc
```

klauer:
Let's re-run this after #557 is in.

klauer:
Both travis/azure are mostly happy, with a few random failures (_oh, caproto..._).

Thanks @MikeHart85, merging now...

PR #574, opened 2020-03-25
DEV: add an optional pre-commit configuration step for development purposes

klauer:
```bash
$ brew install pre-commit   # pip install pre-commit
$ cd /path/to/caproto
$ pre-commit install

# run on all files
$ pre-commit run --all-files
# or run just on changed files
$ echo 'blah' >> setup.py
$ git commit -am "test"   # fails due to linting
```

We've found that this can save a lot of time during development. Only a suggested possible configuration.

Could also rely on isort:
```
-   repo: https://github.com/timothycrosley/isort
    rev: 4.3.21-2
    hooks:
    -   id: isort
```

But this will result in many more changes. If this seems useful, we should still probably wait until 5.0 of isort is released as there are some sorting issues with the current release.


.pre-commit-config.yaml
```diff
@@ -0,0 +1,21 @@
...
+    hooks:
+    -   id: no-commit-to-branch
+    -   id: trailing-whitespace
+    -   id: end-of-file-fixer
+    -   id: check-ast
+    -   id: check-case-conflict
+    -   id: check-merge-conflict
+    -   id: check-symlinks
+    -   id: check-yaml
+        exclude: '^(conda-recipe/meta.yaml)$'
```
danielballan: It might make sense to also exclude `CAproto.html` (and revert the changes to it) as it is a vendored file that we use for code-generation.

danielballan: Thoughts on this, @klauer? If you feel that we _should_ reformat `CAproto.html` I'm OK with it, but my sense it that we should omit it.

Once that's settled one way or the other, I am in favor of merging this.

klauer: I'm 50/50 - trailing whitespace is annoying, but changing vendored files is probably worse.

I'll exclude it from that check, rebase, and force push shortly.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,21 @@
...
+    rev: v2.5.0
+    hooks:
+    -   id: no-commit-to-branch
+    -   id: trailing-whitespace
+    -   id: end-of-file-fixer
+    -   id: check-ast
+    -   id: check-case-conflict
+    -   id: check-merge-conflict
+    -   id: check-symlinks
+    -   id: check-yaml
```
danielballan, line 14: Do we have any YAML in this project?

klauer: Only CI configuration

danielballan:
Regarding the documentation I suggested in my review, I took the liberty of pushing a commit with some docs.

.pre-commit-config.yaml
danielballan: Oh, right. 🐑

.pre-commit-config.yaml
klauer: Resolved in force-push (--force-with-lease, I'm not a complete monster)

PR #576, opened 2020-03-27
FIX: Long string modifier ('PV$') for STRING/CHAR dtypes

klauer:
This is at least working locally - I _believe_ correctly:

```
$ cainfo mock:E.VAL
mock:E.VAL
    State:            connected
    Host:             192.168.99.1:5064
    Access:           read, write
    Native data type: DBF_STRING
    Request type:     DBR_STRING
    Element count:    1

$ cainfo mock:E.VAL$
mock:E.VAL$
    State:            connected
    Host:             192.168.99.1:5064
    Access:           read, write
    Native data type: DBF_CHAR
    Request type:     DBR_CHAR
    Element count:    14

$ caget -S mock:E.RTYP
mock:E.RTYP stringin

$ caget -S mock:E.RTYP$
mock:E.RTYP$ stringin

$ caget -S mock:E.VAL$
mock:E.VAL$ this is a test
```

And not shown for other fields:

```
$ caget -S mock:C.RTYP
mock:C.RTYP ai

$ caget mock:C.VAL$
Channel connect timed out: 'mock:C.VAL$' not found.
```
cc @baldwinb (please chime in if I'm wrong about any of the above)

### DIFF
--- a/caproto/server/common.py
+++ b/caproto/server/common.py
@@ -456,7 +456,9 @@ class VirtualCircuit:
 
             metadata, data = await db_entry.auth_read(
                 self.client_hostname, self.client_username,
-                data_type, user_address=self.circuit.address)
+                data_type, user_address=self.circuit.address,
+                long_string=chan.name.endswith('$')
+            )
 
             old_version = self.circuit.protocol_version < 13
             if command.data_count > 0 or old_version:
--- a/caproto/server/common.py
+++ b/caproto/server/common.py
@@ -8,7 +8,7 @@ import weakref
 import caproto as ca
 from caproto import (apply_arr_filter, get_environment_variables,
                      RemoteProtocolError, CaprotoKeyError, CaprotoRuntimeError,
-                     CaprotoNetworkError)
+                     CaprotoNetworkError, ChannelType)
 from .._dbr import SubscriptionType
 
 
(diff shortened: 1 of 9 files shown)


caproto/server/common.py
```diff
@@ -408,21 +408,30 @@ def get_db_entry():
...
             else:
 
                 access = db_entry.check_access(self.client_hostname,
                                                self.client_username)
 
+                modifiers = ca.parse_record_field(pvname).modifiers
+                data_type = db_entry.data_type
+                data_count = db_entry.max_length
+                if ca.RecordModifiers.long_string in (modifiers or {}):
+                    if data_type in (ChannelType.STRING, ):
```
danielballan: Should this check whether the _native_ type is STRING and then, if so, find the corresponding CHAR type?

caproto/server/common.py
```diff
@@ -720,8 +729,17 @@ def __getitem__(self, pvname):
                 raise CaprotoKeyError(f'Neither record nor field exists: '
                                       f'{rec_field}')
 
-            # Cache record.FIELD for later usage
-            self.pvdb[rec_field] = inst
+        # Verify the modifiers are usable BEFORE caching rec_field in the pvdb:
+        if ca.RecordModifiers.long_string in (mods or {}):
+            if inst.data_type not in (ChannelType.STRING,
```
danielballan: Same here regarding native type vs specific type.

danielballan:
I think this is the right layer to do this in. Just wondering about generality to other string/char types.

caproto/server/common.py
klauer: I believe `db_entry` (`ChannelData`) `.data_type` will never be a non-native type.

danielballan: Ah, of course.

klauer:
I'm a bit stuck here on data conversion, which is the cause of the failed tests.

Case 1: A `stringin` record in EPICS holding `"34"`
Client requests: DBR_STRING -> `"34"`
Client requests: DBR_CHAR -> `34`  *not* a list of bytes `[3, 4]`

Case 2: A `DBF_STRING` field `.VAL` in EPICS, holding "value"
Client requests `.VAL` as DBR_STRING -> `"value"`
Client requests `.VAL$` as DBR_CHAR -> `[ord(c) for c in "value"]`

In either scenario, in caproto, we store the data internally as a `ChannelString` and indicate its native type as `ChannelType.STRING`. 

In this PR, `.VAL$` still just refers to the `ChannelData` field instance `.VAL` which as I just noted holds a `ChannelType.STRING`. The backend conversion is expected to handle the client request.

However, as in the two cases above, we now have two scenarios where we want `STRING` -> `CHAR` conversions to happen in different ways. Either this indicates that the backend needs refactoring, or this PR needs rethinking.

baldwinb:
Hi Ken,
I am not sure if this is helpful, but I tried to dig out how this was originally written. The appending $ first appeared in epics base 3.14.11 according to its release notes.
A presentation i found said this;
> Appending ‘$’ to the name of a string or link field in a PV name changes the
> reported native type from DBR_STRING to array of DBR_CHAR
> Only works when naming string or link fields
> When omitting the field name .VAL, a dot is required
> Use myStringPV.$ or myStringPV.VAL$
> myStringPV$ is a different record
I don't know if that is currently true though.
In the epics base I think src/db/dbAccess.c function epicsShareAPI is where it picks up it has a $ on it. In case that gives some further ideas that are portable to python.

klauer:
Thanks for the input, @baldwinb! I hadn't ever seen `myStringPV.$` so that's a definite oversight here.

klauer:
Pushed a fix in 0d9b0fb for data conversion, which is ugly and changes the `read` API slightly to handle this differing conversion scenario which I noted above.

I'm not happy about it, but unless @danielballan or @tacaswell has a better solution, this is it.

klauer:
Turns out `PREFIX:Record.$` works, whereas `PREFIX:Record.` does not currently. Pushing another fix, then this will be ready, pending some input from @danielballan and @tacaswell.

PR #578, opened 2020-03-28
Provide Subscription to user callback (in sync client).

danielballan:
In #574 we changed the expected signature for user subscription callbacks from
`f(response)` to `f(sub, response)` in the threading client. This PR
applies the same changes to the sync client, and it updates `caproto-monitor`
accordingly.

I went to apply the same change to the async clients (trio, curio) and found
that they are too far behind for this to be applicable. We still haven't
separate the "create subscription object" and "add callback" steps there.
I think we should wait to worry about that until we unify the clients around
a shared utility module, as we have done for the servers.

So, this PR completes the subscriptions refactor and paves the way for 0.5.0.

### DIFF
--- a/caproto/_utils.py
+++ b/caproto/_utils.py
@@ -5,6 +5,7 @@ import argparse
 import array
 import collections
 import functools
+import inspect
 import os
 import random
 import socket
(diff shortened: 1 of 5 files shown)


danielballan:
Always nice when all the caproto checks pass on the first try!

PR #584, opened 2020-04-07
REF/API: deprecate 'mock_record' in favor of 'record' kwarg for pvproperty

### LINKED ISSUE #583: Consider renaming "mock" records
* Just "record"?
* "Scaffold" record? https://github.com/caproto/caproto/pull/577#pullrequestreview-389456272

Either way, keep back-compat with pvproperty kwarg `mock_record`

klauer:
Closes #583 

* Keeps backward-compatibility
* Warns on usage of `mock_record`
* Blows up if both `mock_record` and `record` are specified
* Blows up if both `record` and `field_spec` are specified. While this is backward-incompatible, it should not have been allowed originally.
    - `caproto/ioc_examples/record.py` was causing test failures because of this


klauer:
Still OK with d91cf8d @tacaswell?

PR #585, opened 2020-04-08
Fix git usage in instructions.

danielballan:


klauer:
It's also possible to do `SKIP=flake8 git commit ...` where `flake8` is the pre-commit id.

PR #586, opened 2020-04-09
FIX: empty select() list is also problematic on OSX with KqueueSelector

### LINKED ISSUE #535: threading.client uses a whole core on my laptop while not been active
I have noticed that my CPU utilization goes up(+25% one core) substantially when I create a very simple threading.client before even connection to any PVs. consumes a lot of CPU while not active.
image

``` python
from caproto.threading.client import Context
ctx = Context()
```

more details here /vstadnytskyi/caproto-sandbox/issues/5

klauer:
Closes #535

Verified locally

### DIFF
--- a/caproto/threading/client.py
+++ b/caproto/threading/client.py
@@ -171,8 +171,8 @@ class SelectorThread:
         self._close_event = threading.Event()
         self.selector = selectors.DefaultSelector()
 
-        if sys.platform == 'win32':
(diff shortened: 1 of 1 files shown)


danielballan:
I guess another approach could be to make this thread wait on an `Event` or something until the first socket is added and then again whenever the number of sockets is 0.

PR #588, opened 2020-04-09
REF: remove dummy socket from threading client

klauer:
Closes #587 

Not thoroughly tested

PR #592, opened 2020-04-11
REF/CLN: selector in threaded client

klauer:
Closes #588 

* Attempt to remove `id_to_socket`

See a way to remove `socket_to_id` as well?

### DIFF
--- a/caproto/threading/client.py
+++ b/caproto/threading/client.py
@@ -178,7 +178,6 @@ class SelectorThread:
 
         self._socket_map_lock = threading.RLock()
         self.objects = weakref.WeakValueDictionary()
-        self.id_to_socket = {}
         self.socket_to_id = {}
 
         self._register_sockets = {}  # {socket: object_id}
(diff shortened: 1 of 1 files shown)


caproto/threading/client.py
```diff
@@ -242,29 +237,27 @@ def remove_socket(self, sock):
...
 
                 for sock, obj_id in self._register_sockets.items():
                     self.selector.register(sock, selectors.EVENT_READ,
                                            data=obj_id)
+                self._socket_count += len(self._register_sockets)
                 self._register_sockets.clear()
 
+            if self._socket_count == 0:
+                if self._register_event.wait(timeout=0.1):
+                    self._register_event.clear()
```
danielballan, line 258: When trio removed the `clear()` method from `Event` I took the lesson to heart that, “Once an Event happens it cannot un-happen,” and that in general it’s better/safer to just make a new `Event`. That said, the alternative here is to put a boolean inside a lock, which seems silly if we can just use `Event` exactly as you have done.

klauer: Yeah, I fiddled around with this for a bit and thought it to be the lesser evil.

klauer:
Haven't found a way to remove `socket_to_id`, but at least this is one `dict` down.

PR #596, opened 2020-05-08
FIX: curio version 1.2

klauer:


klauer:
Tests are passing, but Python is segfaulting on exit (perhaps due to recent ophyd release, I'd imagine).

Some of this couldn't easily be written for the old curio - so unless there's a reason to keep support for an old, experimental pre-1.0 dependency, I'd suggest we pin to `curio>=1.2` from now on.

danielballan:
Let’s see if the new pyepics fixes this....

PR #597, opened 2020-05-09
ENH: add multiprocessing SharedMemory example (Python 3.8+ only)

klauer:
Note
-----
This is Python 3.8 only due to the usage of https://docs.python.org/3/library/multiprocessing.shared_memory.html#module-multiprocessing.shared_memory

I have in the past made attempts with [`multiprocessing.SyncManager`](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.managers.SyncManager) but have found it to be extremely slow when it comes to synchronization primitives.

Docstring / IOC description
----------
A multiprocessing.shared_memory.SharedMemory-backed IOC.

``image`` has a shape (``MAX_SIZE``, ) and is generated in a separate
process running :func:`image_generator`.

A :class:`multiprocessing.Pipe` notifies the main process that a new image
is available in a given shared memory slot, which has some associated
metadata in :class:`ImageSlot`.

This pipe is read out and synchronized with the async framework by way
of :func:`queue_handler_thread`.  This simply receives information from
the pipe and sends it along to an async_lib-defined ThreadsafeQueue.
The pvproperty startup hook for ``image`` waits on the queue, updating
the value and metadata internally and shipping it off to clients by the
usual :meth:`ChannelData.write` mechanism.

    Vectors PVs
    -----------
    shm:image (int)

### DIFF
--- a/caproto/ioc_examples/shared_memory.py
+++ b/caproto/ioc_examples/shared_memory.py
@@ -1,0 +1,220 @@
+#!/usr/bin/env python3
+
+import ctypes
+import collections
+import multiprocessing
+import multiprocessing.managers
+import multiprocessing.shared_memory
+import textwrap
+import threading
+import time
+
+import numpy as np
+
+from caproto.server import pvproperty, PVGroup, ioc_arg_parser, run
+
+
+UPDATE_PERIOD_SEC = 0.001
+IMAGE_DTYPE = ctypes.c_uint8
+MAX_SIZE = 1024 * 1024 * 3
+MAX_BYTE_SIZE = MAX_SIZE * ctypes.sizeof(IMAGE_DTYPE)
+NUM_SLOTS = 5
+
+
+class ImageSlot(ctypes.Structure):
+    _fields_ = [
+        ('in_use', ctypes.c_bool),
+        ('timestamp', ctypes.c_double),
+        # ('image', np.ndarray; here in shared memory)
+    ]
+
+
+def find_free_slot(slots):
+    """
+    Find a free slot to update given a list of :class:`ImageSlot`.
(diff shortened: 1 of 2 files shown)


klauer:
Limited to Python 3.8 as indicated above, this is not present in the test suite. Test failures are completely unrelated.

danielballan:
This seems cool. Been too busy getting tutorial content ready to try it locally yet. I think it would be fine to merge 3.8-only examples marked as such. Reopen?

klauer:
Will reopen when test suite gets back to a working state

leofang:
If you don't mind supporting UNIX-like systems only (Linux & OS X), you can use `posix_ipc` (https://pypi.org/project/posix_ipc/) so that it works for Python <3.8. Python 3.8's SharedMemory is based on that.

leofang:
ref: NSLS-II/ptycho_gui#71

klauer:
Thanks for the information, @leofang. I think using the Python standard library here is more in line with caproto's way of doing things, so I'll keep it as-is regardless of the 3.8 requirement.

danielballan:
I guess this is merge-able then?

klauer:
Given my reservations in #616, I'd say this is probably going in the wrong direction.

klauer:
Moved to 'too_clever' (aka `advanced`, aka, `why are you doing this`)

klauer:
_@klauer clicks "Merge pull request"... And the crowd goes mild!_

danielballan:
:clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap:

klauer:
😉 @danielballan

PR #598, opened 2020-05-13
WIP: control the speed of the motor

tacaswell:


caproto/ioc_examples/mini_beamline.py
```diff
@@ -56,8 +57,10 @@ async def mtr(self, instance, async_lib):
     @no_reentry
     async def mtr(self, instance, value):
         disp = (value - instance.value)
+        total_time = disp / self.vel.value
         dwell = .1
-        step_size = .1
+        N = total_time // dwell
+        step_size = disp / N
```
MikeHart85, line 63: This results in a division by zero if `vel`ocity is more than 10 times greater than `disp`lacement.

Has to be *more* than 10 times; exactly 10 times still works fine. The reason for 10 is 1 / `dwell`.

MikeHart85: Possible solution, change line 62 to:

```python
    N = max(1, int(total_time / dwell))
```

I think that also makes line 64 obsolete, so it could be removed.

The reason for switching to explicit cast is because `//` rounds the same way, but returns a float type, which won't be accepted by `range`.

tacaswell: Given that this is a pretty low-fidelity motor simulation, if we would need less than 1 step, we should just return and skip the sleeping.

tacaswell:
Tested by hand.

Issue #548, opened 2020-01-17
Address deprecation of trio.run_sync_in_worker_thread


danielballan:
```
  /home/travis/build/caproto/caproto/caproto/tests/epics_test_utils.py:58: TrioDeprecationWarning: trio.run_sync_in_worker_thread is deprecated since Trio 0.12.0; use trio.to_thread.run_sync instead (https://github.com/python-trio/trio/issues/810)
    raw_stdout, raw_stderr = await trio.run_sync_in_worker_thread(runner)
```

Issue #505, opened 2019-08-12
Address deprecation of trio.Event.clear()


danielballan:
```
TrioDeprecationWarning: trio.Event.clear is deprecated since Trio 0.12.0; use multiple Event objects or other synchronization primitives instead
```

https://github.com/python-trio/trio/issues/637

danielballan:
Closed by #600

Issue #504, opened 2019-08-12
Address deprecation of BlockingTrioPortal


danielballan:
See https://trio.readthedocs.io/en/stable/history.html#trio-0-12-0-2019-07-31

PR #600, opened 2020-06-07
Deprecated features of `trio`

dmgav:
The changes in this PR are addressing deprecated features of `trio`.

- `trio.run_sync_in_worker_thread` replaced by `trio.to_thread.run_sync`.

- `trio.BlockingTrioPortal()` followed by `portal.run(...)` or `portal.run_sync(...)` replaced by `trio.from_thread.run()` or `trio.from_thread.run_sync(...)`

- There is no straightforward fix for deprecated `trio.Event.clear()`. The class `Event` in  `caproto/trio/server.py` was expanded to emulate the original `trio.Event.clear()` method.

The PR is addressing issues https://github.com/caproto/caproto/issues/548, https://github.com/caproto/caproto/issues/505 and https://github.com/caproto/caproto/issues/504.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,5 +11,7 @@ doc/build
 .asv/
 pytest_bench.json
 .pytest_cache/
+.idea/
+.vscode/
 build/
 dist/
--- a/caproto/_utils.py
+++ b/caproto/_utils.py
@@ -75,7 +75,7 @@ __all__ = (  # noqa F822
     'RecordAndField',
     'ThreadsafeCounter',
     '__version__',
-     # sentinels dynamically defined and added to globals() below
+    # sentinels dynamically defined and added to globals() below
     'CLIENT', 'SERVER', 'RESPONSE', 'REQUEST', 'NEED_DATA',
     'SEND_SEARCH_REQUEST', 'AWAIT_SEARCH_RESPONSE',
     'SEND_SEARCH_RESPONSE', 'SEND_VERSION_REQUEST',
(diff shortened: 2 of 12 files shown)


dmgav:
@danielballan I don't completely understand the problem with the original `Event.clear()`. In the proposed workaround, the `clear()` method is creating a new event if the event is set and it does nothing if the event is not set. It doesn't seem like replacing `trio.Event` with `trio.Condition` will change anything in the context it is used.

danielballan:
The best primitive for this case might be a read--write lock. Documenting what @tacaswell said on the call:

> The task that is trying to write a value would grab the write lock and everything else would grab a read lock.

See the discussion that lead to [this resolution](https://github.com/python-trio/trio/issues/637#issuecomment-499883676).

Links to read-write lock implementations. Note that caproto goes out of its way to have all its core functionality with zero dependencies, so we would want to consider vendoring this if the code is small enough.

* https://github.com/aio-libs/aiorwlock
* https://github.com/python-trio/trio/issues/528

danielballan:
Power-cycled to re-run against pyepics 3.4.2, just released.

danielballan:
Although read--write lock might be conceptually better and result in code that is easier to reason about, none of the three async libraries we use currently provide one. Therefore, I think we should stick with @dmgav's implementation at least for now. I'll open a new issue for exploring read--write locks, as an academic exercise.

danielballan:
Some `flake8` bit rot sneaked in during the period when pyepics broke our CI. Addressed in 1aa3e08. I'll wait for CI to pass again before merging.

@klauer FYI, it looks like `flake8` now takes issue with your policy of "Fine, I'll just put `f` on _all_ my strings including strings like `f''` just in case it comes up!" :- D

danielballan:
One random failure. Good to go.

PR #612, opened 2020-06-18
Avoid duplicate registration of callbacks.

### LINKED ISSUE #607: connection_state_callback registered twice in `caproto.threading.client`
`caproto.threading.client.Context().get_pvs()` registers a `connection_state_callback` twice if the PV instance is newly created.

### Steps to Reproduce

In one terminal start an example IOC:
```bash
python -m caproto.ioc_examples.simple --list-pvs
```
Now, when running the following Python code, the callback `connection()` is called twice:
```python
import caproto, caproto.threading

ctx = caproto.threading.client.Context()

def connection(pv, state):
    print(f"### {pv.name} {state}")

simple_A = ctx.get_pvs("simple:A", connection_state_callback=connection)
```
as the output reveals:
```
### simple:A connected
### simple:A connected
```

**Expected behaviour:**
Above output only once.

### Analysis and Potential Fix

The problem seems to be that the in `Content.get_pvs()` PV objects are [instantiated with the callback `connection_state_callback` in the __init__() call](https://github.com/caproto/caproto/blob/v0.5.1b1/caproto/threading/client.py#L1039) and a few lines later, the callback is added again manually using [`add_callback()`](https://github.com/caproto/caproto/blob/v0.5.1b1/caproto/threading/client.py#L1045).

A potential fix:

```diff
diff --git a/caproto/threading/client.py b/caproto/threading/client.py
index f6cd19a..46036f4 100644
--- a/caproto/threading/client.py
+++ b/caproto/threading/client.py
@@ -1036,8 +1036,7 @@ class Context:
                 try:
                     pv = self.pvs[(name, priority)]
                 except KeyError:
-                    pv = PV(name, priority, self, connection_state_callback,
-                            access_rights_callback, timeout)
+                    pv = PV(name, priority, self, None, None, timeout)
                     names_to_search.append(name)
                     self.pvs[(name, priority)] = pv
                     self.pvs_needing_circuits[name].add(pv)
```

or:

```diff
diff --git a/caproto/threading/client.py b/caproto/threading/client.py
index f6cd19a..17fda79 100644
--- a/caproto/threading/client.py
+++ b/caproto/threading/client.py
@@ -1035,20 +1035,18 @@ class Context:
             with self.pv_cache_lock:
                 try:
                     pv = self.pvs[(name, priority)]
+                    if connection_state_callback is not None:
+                        pv.connection_state_callback.add_callback(
+                            connection_state_callback, run=True)
+                    if access_rights_callback is not None:
+                        pv.access_rights_callback.add_callback(
+                            access_rights_callback, run=True)
                 except KeyError:
                     pv = PV(name, priority, self, connection_state_callback,
                             access_rights_callback, timeout)
                     names_to_search.append(name)
                     self.pvs[(name, priority)] = pv
                     self.pvs_needing_circuits[name].add(pv)
-
-            if connection_state_callback is not None:
-                pv.connection_state_callback.add_callback(
-                    connection_state_callback, run=True)
-            if access_rights_callback is not None:
-                pv.access_rights_callback.add_callback(
-                    access_rights_callback, run=True)
-
             pvs.append(pv)
```

Which of the fixes is better? I don't know. Maybe the second option is better if it is important that the callback is already registered, right when the object is instantiated (eg. if a state change could be missed otherwise). This is for you to judge.

danielballan:
Closes #607

Thanks again, @pklaus!

### DIFF
--- a/caproto/tests/test_threading_client.py
+++ b/caproto/tests/test_threading_client.py
@@ -777,8 +777,12 @@ def test_registration_of_connection_state_callback(context, ioc):
     def cb(pv, state):
         ...
 
-    pv, = context.get_pvs(ioc.pvs['str'], connection_state_callback=cb)
+    pv, = context.get_pvs(
+        ioc.pvs['str'],
+        connection_state_callback=cb,
+        access_rights_callback=cb)
     assert len(pv.connection_state_callback.callbacks) == 1
+    assert len(pv.access_rights_callback.callbacks) == 1
 
 
 def test_time_since_last_heard(context, ioc):
--- a/caproto/tests/test_threading_client.py
+++ b/caproto/tests/test_threading_client.py
@@ -770,6 +770,17 @@ def test_events_off_and_on(ioc, context):
     assert monitor_values[1:] == [1, 2, 3, 6, 7, 8, 9]
 
 
+def test_registration_of_connection_state_callback(context, ioc):
+    # Check for bug caught in https://github.com/caproto/caproto/issues/607
+    # wherein callback was registered twice.
+
+    def cb(pv, state):
+        ...
+
+    pv, = context.get_pvs(ioc.pvs['str'], connection_state_callback=cb)
+    assert len(pv.connection_state_callback.callbacks) == 1
+
+
 def test_time_since_last_heard(context, ioc):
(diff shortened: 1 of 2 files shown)


danielballan:
One random failure on one build.

PR #614, opened 2020-07-01
API: bump min Python to 3.7

tacaswell:
Consistent with NEP29

This will also make implementing the asyncio client easier.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,18 +1,13 @@
 language: python
-
+os: linux
 dist: xenial
-
-sudo:
-  - false
-  - true
-
 
 python:
   - 3.7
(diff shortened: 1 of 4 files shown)


tacaswell:
https://config.travis-ci.com/explore is a magical tool!

danielballan:
Unless we have a strong reason to need 3.7, I would be happy to let caproto core support 3.6 for awhile longer. We don't really have the bandwidth for maintenance branches of caproto, so once it's dropped, it's dropped, and continued support for 3.6  (at least, in our core functionality) might help with adoption.

klauer:
Superseded by #794 - Python 3.8+

Issue #579, opened 2020-04-01
IOC example that contains an Epics client?


rajanplumley:
Can an example be made that demonstrates how one might use an Epics client in conjunction with the caproto server to read and write to external PVs?

tacaswell:
The problem is that all of the IOCs are async based but the only clients we have finished writing are the threaded and sync clients.

I _think_ you can just use the threaded client in the IOC, but it would be much better to get the async clients written....

danielballan:
Hello, @rajanplumley! This is a good suggestion, one I think we can act on, though it raises an issue I will comment on at length below.

Ideally, one would use a single concurrency framework (threads, asycnio, trio, curio, etc.) for both the server and client aspects of the program. Unfortunately, our servers use asyncio or trio or curio, and our high-quality client uses threads. (We also have a synchronous client that is of high quality, but its synchronous nature makes it unsuitable for use inside a server.) Therefore, at the moment, such an example would have to use asyncio/trio/curio for the server and threads for the client aspects, which is a bit weird, but certainly workable. We have _sketches_ of trio and curio clients. In the future we hope to bring them up to the same level of quality as the threading-based client, and also to add an asyncio-based client.

All that said, a working server that uses the threading client inside would be useful as a stop-gap, even if it would be a bit of a Frankenstein.

danielballan:
Ha, _jinx_ @tacaswell. :-D

rajanplumley:
Thanks @danielballan and @tacaswell.  I'll experiment with using the threading client inside the server and report any surprises ;)

klauer:
Added the example, which is now merged. 
Hopefully we'll get to an asyncio client example one day soon as well.

PR #615, opened 2020-07-06
ENH: add IOC example for monitoring and mirroring an external PV

klauer:
~- Future: Create a generalized helper method to do this setup for you~
 - Future: allow customization of PV to be mirrored (fixed at "simple:A" for now)
 - Could be simplified if link status does not need to be replicated

Related https://github.com/caproto/caproto/issues/579 (does not close the issue)

cc @rajanplumley

### DIFF
--- a/caproto/ioc_examples/thread_client_monitor.py
+++ b/caproto/ioc_examples/thread_client_monitor.py
@@ -7,6 +7,110 @@ from caproto.server import pvproperty, PVGroup, ioc_arg_parser, run
 from textwrap import dedent
 
 
+_default_thread_context = None
+
+
+def get_default_thread_context():
+    """Get a shared caproto threading client context."""
+    global _default_thread_context
+    if _default_thread_context is None:
+        _default_thread_context = caproto.threading.client.Context()
+    return _default_thread_context
+
+
+def _monitor_pvs(*pv_names, context, queue, data_type='time'):
+    """
+    Monitor pv_names in the given threading context, putting events to `queue`.
+
+    Parameters
+    ----------
+    *pv_names : str
+        PV names to monitor.
+
+    context : caproto.threading.client.Context
+        The threading context to use.
+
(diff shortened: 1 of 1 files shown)


caproto/ioc_examples/thread_client_monitor.py
```diff
@@ -0,0 +1,168 @@
...
+
+    Returns
+    -------
+    subscriptions : list
+        List of subscription tuples, with each being:
+        ``(sub, subscription_token, *callback_references)``
+    """
+
+    def add_to_queue(sub, event_add_response):
+        queue.put(('subscription', sub.pv, event_add_response))
```
danielballan, line 47: Might as well expose `sub` here, maybe. Not used downstream in this case, but it could be in future variations.

caproto/ioc_examples/thread_client_monitor.py
```diff
@@ -0,0 +1,168 @@
...
+    if context is None:
+        context = get_default_thread_context()
+
+    queue = async_lib.ThreadsafeQueue()
+    subscriptions = _monitor_pvs(*pv_names, context=context, queue=queue,
+                                 data_type=data_type)
+    try:
+        while True:
+            info = await queue.async_get()
+            yield info
```
danielballan, line 108: I got tripped up briefly by the fact that the local variable `info` is used for the whole bundle here on the send side and used on the receive side below to mean just the `event_add_response`.

caproto/ioc_examples/thread_client_monitor.py
klauer: Addressed in d625a2f

caproto/ioc_examples/thread_client_monitor.py
klauer: Addressed in d625a2f

danielballan:
Kicked failed builds. Merge at will if CI passes.

PR #617, opened 2020-07-27
ENH: add proof-of-concept autosave tools

### LINKED ISSUE #546: Introduce autosave-like functionality in the server

klauer:
Closes #546 

Usage

```python
    At the top-level PVGroup, ensure only one AutosaveHelper SubGroup has been
    added. Then, wrap any pvproperty to be saved with ``autosaved``:

    ::

        autosave_helper = SubGroup(AutosaveHelper)
        value = autosaved(pvproperty(1.0))
```

* Save period and filename can be modified in `YourPVGroup.__init__`.
* Saved in convenient JSON format (which is notably not compatible with the standard EPICS autosave tools)
* Fields can likewise be saved (by attribute name, at least for now) by specifying `autosaved(..., fields=['description', ...])`; description is done by default.
* Autosave-tagged pvproperties in subgroups will be found, so only the top-level AutosaveHelper is necessary.
* `PVSpec` is inflexible/immutable by design, so a small hack around that was necessary here. If this is built into caproto, we can at least rework that hack later.
* Custom file rotating tool that deletes old files or ones over a maximum amount per day

### DIFF
--- a/caproto/server/autosave.py
+++ b/caproto/server/autosave.py
@@ -1,17 +1,202 @@
 import json
 import logging
-import logging.handlers
 import pathlib
+import datetime
+import operator
+import tempfile
+import typing
 
-from . import PVGroup, PVSpec, pvproperty
+from . import PVGroup, pvproperty
 
 
-def get_autosave_fields(pvprop):
+MODULE_LOGGER = logging.getLogger(__name__)
+
+
+class RotatingFileManager:
+    """
+    Save to a primary filename, while rotating out old copies.
+
+    Automatically prunes old copies of the file beyond a certain age or a
+    maximum file count (overall or per day).
+
+    Files will first be pruned by the overall age cutoff, then on a per day
+    basis.
+
+    Parameters
+    ----------
+    filename : str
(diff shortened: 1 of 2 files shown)


caproto/server/autosave.py
```diff
@@ -0,0 +1,175 @@
...
+    def _replace(self, **kwargs):
+        # NOTE: a further hack: this is a feature from `namedtuple` that's used
+        # to replace item(s) in the tuple, returning a newly modified copy.
+        # Here, we attach on `autosave` settings after performing the replace.
+        replaced = super()._replace(**kwargs)
+        replaced.autosave = self.autosave
+        return replaced
+
+
+def autosaved(pvprop, fields=None):
```
danielballan, line 139: Is there a way to make this fail loudly and helpfully if there is no `autosave_hook` running?

klauer: Hmm, maybe. Good idea - I think this could be an issue for later.

danielballan:
One random CI failure. Merging.

PR #618, opened 2020-07-28
FIX: missing f-string in server log

klauer:
Simple fix.

@danielballan @ke-zhang-rd do you have a recommendation as to best setting up logging with an IOC external to caproto? e.g., a package `my_ioc` has `MyPVGroup`, making for a logger name `my_ioc.MyPVGroup`.

I'm thinking I'd like to use the same handler to get the same record formatting and save to the same file (perhaps even tee to standard output as well). It doesn't seem to fit well into the current structure, but I may be missing something.

### DIFF
--- a/caproto/server/server.py
+++ b/caproto/server/server.py
@@ -1220,7 +1220,7 @@ class PVGroup(metaclass=PVGroupMeta):
             base = self.__class__.__module__
 
         # Instantiate the logger
-        self.log = logging.getLogger('{base}.{log_name}')
+        self.log = logging.getLogger(f'{base}.{log_name}')
         self._create_pvdb()
 
         # Prime the snapshots to the current state.


tacaswell:
:+1: to self-merging on a reasonable number of the tests passing.

danielballan:
Could you do

```py
from caproto import get_handler

logging.getLogger('my_ioc.MyGroup').addHandler(get_handler())
```

klauer:
@danielballan maybe, but I haven't found the right combination yet due to several things.

* Server `run()` never actually calls `config_caproto_logging()`: https://github.com/caproto/caproto/blob/65eb6367a97f24c99bcf66603ebb8d9dccbec338/caproto/server/server.py#L1378-L1387
* The above means that `get_handler()` returns `None` unless - somewhat bizarrely - in quiet or verbose modes due to the `if` block logic.
* It's awkward when the desired levels don't match (e.g., give me caproto warning logs but IOC debug logs)

This semi-works but duplicates `ctx` messages because of the above set_handler code:

```python
    ioc = create_ioc(**ioc_args, **ioc_options)

    caproto.config_caproto_logging()  #  necessary to get the handler
    handler = caproto.get_handler()
    handler.setLevel('DEBUG')   # we want our server's debug messages
    ioc.log.setLevel('DEBUG')   # this needs to be at DEBUG so we can see server debug messages
    ioc.log.addHandler(handler)

    run(ioc.pvdb, **run_options)
```

e.g.,
```
[I 11:51:29.926       server:  165] Asyncio server starting up...
[I 11:51:29.926       server:  165] Asyncio server starting up...
[I 11:51:29.926       server:  177] Listening on 0.0.0.0:62645
[I 11:51:29.926       server:  177] Listening on 0.0.0.0:62645
```

Next best thing was this, but it separates the streams:
```
    caproto_log._set_handler_with_logger(logger_name=ioc.log.name)
```

Separating the two may be desirable, but may not be depending on what the user wants.

klauer:
Random test failures. Merging.

PR #627, opened 2020-07-30
FIX: asyncio queue put must use run_coroutine_threadsafe

klauer:
I think this is a reasonable trade-off from the previous implementation: you must create the queue in an async function so it knows which loop to use.

If not, we can return to the 
```
def create_queue_class(loop):
   class AsyncioQueue:
     ...
```
style we had before.


klauer:
Unrelated failure in curio client tests. Merging.

Issue #622, opened 2020-07-28
External package IOC server logging best practices


klauer:
Bringing this issue out from a PR. Original messages:

```py
from caproto import get_handler

logging.getLogger('my_ioc.MyGroup').addHandler(get_handler())
```

_Originally posted by @danielballan in https://github.com/caproto/caproto/pull/618

* Server `run()` never actually calls `config_caproto_logging()`: https://github.com/caproto/caproto/blob/65eb6367a97f24c99bcf66603ebb8d9dccbec338/caproto/server/server.py#L1378-L1387
* The above means that `get_handler()` returns `None` unless - somewhat bizarrely - in quiet or verbose modes due to the `if` block logic.
* It's awkward when the desired levels don't match (e.g., give me caproto warning logs but IOC debug logs)

This semi-works but duplicates `ctx` messages because of the above set_handler code:

```python
    ioc = create_ioc(**ioc_args, **ioc_options)

    caproto.config_caproto_logging()  #  necessary to get the handler
    handler = caproto.get_handler()
    handler.setLevel('DEBUG')   # we want our server's debug messages
    ioc.log.setLevel('DEBUG')   # this needs to be at DEBUG so we can see server debug messages
    ioc.log.addHandler(handler)

    run(ioc.pvdb, **run_options)
```

e.g.,
```
[I 11:51:29.926       server:  165] Asyncio server starting up...
[I 11:51:29.926       server:  165] Asyncio server starting up...
[I 11:51:29.926       server:  177] Listening on 0.0.0.0:62645
[I 11:51:29.926       server:  177] Listening on 0.0.0.0:62645
```

Next best thing was this, but it separates the streams:
```
    caproto_log._set_handler_with_logger(logger_name=ioc.log.name)
```

Separating the two may be desirable, but may not be depending on what the user wants.

_Originally posted by @klauer in https://github.com/caproto/caproto/pull/618#issuecomment-665216853_

klauer:
I hope to add recommended logging practices for IOCs to the cookiecutter in the near future - but still no clear path forward here to me.

PR #628, opened 2020-07-31
DOC: Release notes for 0.6.0

klauer:
Didn't discuss this prior to choosing a version number or any details otherwise, but here we go regardless.

Would have been nice to figure out https://github.com/caproto/caproto/issues/622 but this can come in the next minor release, I suppose.

Adding doctr-versions-menu support, unless anyone objects to doing so. LCLS is using it effectively recently without much issue.


klauer:
One trio client failure. Merging.

PR #633, opened 2020-08-01
DOC: bump copyright date, add rebuild script

klauer:
This is the result of my local rebuilding efforts to get doctr-versions-menu going.

I'm cautiously optimistic that new tags / master rebuilds won't break it, but we'll see.

Note: per-version docs are already available as a result of this. It's not required to merge this for that purpose.
https://caproto.github.io/caproto/master/
See the version selector on the bottom left:
image

### DIFF
--- a/doc/rebuild.sh
+++ b/doc/rebuild.sh
@@ -1,0 +1,27 @@
+#!/bin/bash
+
+GITHUB_IO=../github.io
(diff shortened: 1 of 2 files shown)


klauer:
CI irrelevant

danielballan:
Right 🐑

PR #637, opened 2020-08-14
Add 'WaitForPlugins' to spoof_beamline IOC.

danielballan:


danielballan:
Trivial changes to just one example IOC, useful for beamline simulation. Self-mering.

PR #643, opened 2020-08-26
REF/FIX: use new `get_client_address_list`; do not use hard-coded ports

### LINKED ISSUE #639: Caproto adds additional port to EPICS address
Using `caget` on the machine works, but `caproto-get` gives the following error:
```
caproto._utils.CaprotoNetworkError: Failed to send to xxx.xxx.xxx.xxx:5066:5065
```
It looks like caproto is picking up the correct address and port from `EPICS_CA_ADDR_LIST` but then appends another (the default?) port.

### LINKED ISSUE #645: Client repeater registration logic needs review
While the logic for binding + starting the repeater seems correct, it appears that caproto incorrectly attempts to register with all _remote_ addresses in `EPICS_CA_ADDR_LIST`:

https://github.com/caproto/caproto/blob/b1286f1863fde6c21f1092ceab462cd6f2598db5/caproto/sync/client.py#L62-L67

Other clients are similar, though they may not respect `EPICS_CA_REPEATER_PORT`, using `caproto._constants.EPICS_CA2_PORT` instead.

Referring to epics-base, this should be entirely a local operation:
https://github.com/epics-base/epics-base/blob/524ceee2c8ce8f8446c46d93b5a467edd14b8cc5/src/client/udpiiu.cpp#L465-L556

Registering with the repeater by way of:

```c
        saddr = osiLocalAddr ( sock );
```

`osiLocalAddr` inspects the interfaces by way of `socket_ioctl` (`SIOCGIFCONF`), searching for the first one that fits its criteria - falling back to loopback (`INADDR_LOOPBACK`) on failure
(https://github.com/epics-base/epics-base/blob/8ce980f6637d85fb1f5a139bab404269734b31bc/src/osi/os/default/osdNetIntf.c#L251-L345). This local address determination is performed once and cached by way of `epicsThreadOnce`.

Registration attempt sending:
```c
    AlignedWireRef < epicsUInt16 > ( msg.m_cmmd ) = REPEATER_REGISTER;
    (...clipped...)
    status = sendto ( sock, (char *) &msg, len, 0,
                      &saddr.sa, sizeof ( saddr ) );
```

klauer:
* Leaves `ca.get_address_list` behind for (mostly) back-compat
* Adds `get_client_address_list`, which pairs well enough with `get_server_address_list` and `get_beacon_address_list` 
* Removes hard-coded ports in some places

Badly in need of double-checking. ~Will mark draft until I can find the time.~

~Logic in broadcasters particularly could use further refactoring, as we require the broadcaster instance to know which port it wants to send information on, rather than just saying `"send this to the address list with whatever the configured {repeater, ca server port} is"`~

Repeater registration with local address, removing `port` from `broadcaster.send(port)`.
- [x] asyncio client
- [x] threading client
- [x] sync client
- [x] curio, trio clients

Respecting ports by using `caproto.get_client_address_list`
- [x] asyncio client
- [x] threading client
- [x] sync client
- [x] curio, trio clients
- [x] `x:1234` as `EPICS_CA_SERVER_PORT` means to listen on udp/tcp on `1234`; so this is the appropriate port for broadcasts
- [x] PVA uses `EPICS_PVA_BROADCAST_PORT`; though it does not appear that setting `EPICS_PVAS_SERVER_PORT` alone effects the same broadcast port change, instead requiring `EPICS_PVAS_BROADCAST_PORT` (meaning caproto-pva will still be buggy with custom ports)

Closes #645 
Closes #639

### DIFF
--- a/caproto/_utils.py
+++ b/caproto/_utils.py
@@ -351,12 +351,20 @@ def get_client_address_list(*, protocol=Protocol.ChannelAccess):
     Get channel access client address list in the form of (host, port) based on
     environment variables.
 
+    For CA, the default port is ``EPICS_CA_SERVER_PORT``.
+    For PVA, the default port is ``EPICS_PVA_BROADCAST_PORT``.
+
     See Also
     --------
     :func:`get_address_list`
     '''
     protocol = Protocol(protocol)
-    default_port = get_environment_variables()[f'EPICS_{protocol}_SERVER_PORT']
+    env = get_environment_variables()
+    if protocol == Protocol.ChannelAccess:
+        default_port = env['EPICS_CA_SERVER_PORT']
+    else:
+        default_port = env['EPICS_PVA_BROADCAST_PORT']
+
     return list(set(
         get_address_and_port_from_string(addr, default_port)
         for addr in get_address_list(protocol=protocol)
--- a/caproto/_utils.py
+++ b/caproto/_utils.py
@@ -346,7 +346,7 @@ def get_address_list(*, protocol=Protocol.ChannelAccess):
     return list(set(addresses + auto_addr_list))
 
 
-def get_client_address_list(port=None, *, protocol=Protocol.ChannelAccess):
+def get_client_address_list(*, protocol=Protocol.ChannelAccess):
     '''
     Get channel access client address list in the form of (host, port) based on
     environment variables.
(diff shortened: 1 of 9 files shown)


klauer:
Needs work based on discussion from #645

caproto/_utils.py
```diff
@@ -422,6 +488,64 @@ def get_netifaces_addresses():
                     yield (peer, peer)
 
 
+@functools.lru_cache(maxsize=1)
```
danielballan, line 491: Smart.

caproto/asyncio/client.py
```diff
@@ -116,21 +116,15 @@ async def send(self, port, *commands):
                 'our_address': self.broadcaster.client_address,
                 'direction': '--->>>'}
 
-        for host in ca.get_address_list():
```
danielballan, line 119: Nice to see this oft-repeated code snippet factored out into a utility.

PR #648, opened 2020-09-03
FIX: caproto-get/monitor char array

klauer:
I thought this worked before but frustratingly couldn't find a way to get an array of bytes to format a string.

With this PR, long strings are no problem:
```
$ caproto-get -s AT2L0:CALC:STATS:TraceDiffResults.VAL$
AT2L0:CALC:STATS:TraceDiffResults.VAL$    /Users/klauer/mc/envs/hxr-attenuator/lib/python3.7/tracemalloc.py:65: size=70.5 KiB (-70.5 KiB), count=902 (-902), average=80 B
```

Unclear if this will have any side effects - I don't really understand much of the CLI formatting code.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -21,7 +21,7 @@ repos:
     -   id: flake8
 
 -   repo: https://github.com/timothycrosley/isort
-    rev: 5.4.2
+    rev: 5.5.0
     hooks:
     -   id: isort
 
--- a/caproto/commandline/cli_print_formats.py
+++ b/caproto/commandline/cli_print_formats.py
@@ -5,8 +5,8 @@ This module contains functions for formatting data output for CLI utilities
 '''
 
 import numbers
-import sys
 import re
+import sys
 
 
 class _DataPrintFormat:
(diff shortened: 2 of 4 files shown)


dmgav:
Interesting. Actually an array of bytes should be handled by this condition: https://github.com/caproto/caproto/blob/c9c49f5d69e8e7c3fda0e909348a9bc1204bdcb5/caproto/commandline/cli_print_formats.py#L132 
I guess the problem could be that the string is represented as an array of int? There is a separate option `-S` to print array of char as a string which is not implemented in caproto. The option `-s` seems to be for the cases when a floating point number is sent as a string. I will take a look later today.

klauer:
Yes, array of int representing a string needs to be handled with our tools here. It's a very common method for long strings beyond 40 character DBR_STRING fields.

dmgav:
I think in this case the best approach would be to implement `-S` (captial S) option. This would be consistent with options of `caget`. The code in this PR shouldn't cause any side-effects if invoked with the parameter `-S`.

klauer:
Would you by chance have the time to implement it the proper way with `-S` @dmgav?

dmgav:
@klauer Yes, I will do it.

dmgav:
@klauer I added a new parameter `-S` to both `get` and `monitor`, which sets `args.array_as_string`, which in turn is used to determine if array of int should be converted to a string. It shouldn't change how the rest of the options behave. Can you test if everything works as expected?

klauer:
Thanks, @dmgav. This appears to work. Waiting on tests.

PR #649, opened 2020-09-04
ENH: add IOC stats-like informational tool

klauer:
Continuing building out the "caproto standard library" of sorts - things you'd expect to find on nearly every real EPICS IOC, like autosave. This adds some status information.

Capitalized PV names are standardized names, so I'd prefer to keep those.

- [x] User-togglable tracemalloc debugging of allocations, shipped over CA
- [x] CPU usage
- [x] File descriptor usage
- [x] IOC PID and parent PID
- [x] File descriptor limits (unix)
- [x] PV count
- [x] Startup time
- [x] Heartbeat
- [x] Number of times the IOC has started (requires autosave)
- [x] IOC restart helper
- [x] Directory information
- [x] Number of CPUs
- [x] OS/kernel information
- [x] Caproto / python version info
- [x] Disable tracemalloc on startup

Simply instantiate as subgroup:

```
from caproto.server import PVGroup, SubGroup
from caproto.server.stats import StatusHelper

class MyIOC(PVGroup):
    stats = SubGroup(StatusHelper, prefix=':STATS:')
```

<details>
<summary>Example</summary>

```
PREFIX:STATS:ACCESS        Running
PREFIX:STATS:STARTTOD 2020-09-03 17:36:55.215026
PREFIX:STATS:TOD 2020-09-03 17:37:01.228745
PREFIX:STATS:HEARTBEAT     1
PREFIX:STATS:START_CNT     39
PREFIX:STATS:SYSRESET      0
PREFIX:STATS:RECORD_CNT    181
PREFIX:STATS:FD_MAX        256
PREFIX:STATS:FD_CNT        0
PREFIX:STATS:FD_FREE       0
PREFIX:STATS:SYS_CPU_LOAD  0
PREFIX:STATS:IOC_CPU_LOAD  0
PREFIX:STATS:SUSP_TASK_CNT 0
PREFIX:STATS:MEM_USED      440
PREFIX:STATS:MEM_FREE      0
PREFIX:STATS:MEM_MAX       0
PREFIX:STATS:UPTIME        0
PREFIX:STATS:HOSTNAME      0
PREFIX:STATS:APP_DIR
PREFIX:STATS:ENGINEER      0
PREFIX:STATS:LOCATION      0
PREFIX:STATS:PROCESS_ID    60916
PREFIX:STATS:PARENT_ID     55410
PREFIX:STATS:CPU_CNT       0
PREFIX:STATS:KERNEL_VERS   68
PREFIX:STATS:EPICS_VERSION 0.6.0+29.g2e282b10 on 3.7.6 | packaged by conda-forge | (default, Jun  1 2020, 18:33:30) \n[Clang 9.0.1 ]
PREFIX:STATS:UPD_TIME      15


$ caproto-get -s PREFIX:STATS:TraceTopAllocations
PREFIX:STATS:TraceTopAllocations      #1: /Users/klauer/mc/envs/hxr-attenuator/lib/python3.7/linecache.py:137: 651.7 KiB
    lines = fp.readlines()
#2: /Users/klauer/mc/envs/hxr-attenuator/lib/python3.7/tracemalloc.py:185: 147.6 KiB
    self._frames = tuple(reversed(frames))
#3: /Users/klauer/mc/envs/hxr-attenuator/lib/python3.7/threading.py:238: 115.4 KiB
    self._waiters = _deque()
#4: /Users/klauer/mc/envs/hxr-attenuator/lib/python3.7/tracemalloc.py:65: 113.9 KiB
    return (self.size, self.count, self.traceback)
#5: /Users/klauer/mc/envs/hxr-attenuator/lib/python3.7/tracemalloc.py:113: 70.2 KiB
    self.traceback)
#6: <frozen importlib._bootstrap_external>:525: 52.8 KiB
#7: /Users/klauer/docs/Repos/caproto/caproto/_commands.py:377: 49.8 KiB
    return data[total_size:], command, 0
#8: /Users/klauer/mc/envs/hxr-attenuator/lib/python3.7/threading.py:86: 47.4 KiB
    return _CRLock(*args, **kwargs)
#9: /Users/klauer/mc/envs/hxr-attenuator/lib/python3.7/threading.py:500: 41.8 KiB
    self._cond = Condition(Lock())
#10: /Users/klauer/docs/Repos/caproto/caproto/threading/client.py:2027: 36.8 KiB
    self.callbacks = {}


```
</details>

### DIFF
--- a/caproto/server/stats.py
+++ b/caproto/server/stats.py
@@ -14,36 +14,10 @@ from .autosave import autosaved
 MODULE_LOGGER = logging.getLogger(__name__)
 
 
-def get_top_allocation_info(snapshot: tracemalloc.Snapshot, *,
-                            key_type: str = 'lineno',
-                            limit: int = 10
-                            ) -> typing.List[str]:
-    snapshot = snapshot.filter_traces((
-        tracemalloc.Filter(False, "<frozen importlib._bootstrap>"),
-        tracemalloc.Filter(False, "<unknown>"),
-        tracemalloc.Filter(False, "<tracemalloc>"),
-    ))
-    top_stats = snapshot.statistics(key_type)
-
-    result = []
-    for index, stat in enumerate(top_stats[:limit], 1):
-        frame = stat.traceback[0]
-        kbytes = stat.size / 1024
-        result.append(
-            f"#{index}: {frame.filename}:{frame.lineno}: {kbytes:.1f} KiB"
-        )
-        line = linecache.getline(frame.filename, frame.lineno).strip()
-        if line:
-            result.append(f'    {line}')
-
-    return result, top_stats
-
-
 class StatusHelper(PVGroup):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self._root_pvgroup = self._find_top_level_pvgroup()
-        self._old_snapshot = None
 
     def _find_top_level_pvgroup(self):
         ancestor = self.parent
@@ -330,9 +304,21 @@ class StatusHelper(PVGroup):
     )
 
     async def _update_status_periodic(self):
-        if self._root_pvgroup is None:
-            return
-        await self.record_count.write(value=len(self._root_pvgroup.pvdb))
+        import os
+        try:
+            import psutil
+        except ImportError:
+            self.log.warning(
+                "The Python library psutil is not installed, so MEM_USED will "
+                "not be reported by StatsHelper.")
+            process = psutil.Process(os.getpid())
+            skip_mem_used = True
+        else:
+            skip_mem_used = False
+        if self._root_pvgroup is not None:
+            await self.record_count.write(value=len(self._root_pvgroup.pvdb))
+        if not skip_mem_used:
+            await self.mem_used.write(value=process.memory_info().rss)
 
     @update_period.startup
     async def update_period(self, instance, async_lib):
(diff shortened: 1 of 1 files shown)


caproto/server/stats.py
```diff
@@ -76,20 +82,24 @@ def __init__(self, *args, **kwargs):
 
     engineer = pvproperty(
         value='',
+        max_length=40,
         name='ENGINEER',
         record='stringin',
         read_only=True,
+        doc='Who is responsible this abomination',
```
danielballan, line 96:
```suggestion
        doc='Who is responsible for this abomination',
```
I only review jokes.

klauer: Just under 40 chars; I'll accept it

caproto/server/stats.py
```diff
@@ -0,0 +1,562 @@
...
+    )
+    autosaved(start_count)
+
+    @start_count.startup
+    async def start_count(self, instance, async_lib):
+        # Give autosave some time to load
+        await async_lib.library.sleep(3)
+        await self.start_count.write(value=self.start_count.value + 1)
+
+    # TODO: any way we can find this information?
```
danielballan, line 257: This is tough. The number of connected clients is available via the logging stream, so at least there is a way to know it, but it would require significant changes to thread that into here. `run(...)` consumes the `pvdb` dict and has no way to inject itself in.

In fact, making these objects aware of their server might be abstraction-breaking, as a given `pvdb` could conceivably be run by more than one server. That's a wild and probably useless possibility, but just a way of saying there is currently no 1-to-1 coupling here and implementing this would force us to introduce it.

danielballan (commented):
Looks good. Merge at will once you look at one possible important typo, noted in line.

caproto/server/stats.py
```diff
@@ -0,0 +1,562 @@
...
+        value=0,
+        name='HEARTBEAT',
+        record='calcout',
+        doc='1 Hz counter since startup',
+        read_only=True,
+    )
+
+    @heartbeat.scan(period=1)
+    async def heartbeat(self, instance, async_lib):
+        await self.heartbeat.write(value=1 - self.heartbeat.value)
```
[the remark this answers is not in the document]
klauer: I was going to say "no" here but double-checked. I had assumed this was a 0-1-0-1-style heartbeat, but it is actually as you assumed: https://github.com/slac-epics/iocAdmin/blob/f4b9a558e1933f2ae155b2b579a0fb3e739c554d/iocAdmin/Db/ioc.template#L30-L36

caproto/server/stats.py
klauer: Regarding abstractions, your thoughts mirror my own. I rather like that they are separated, but its effect is unfortunate in this case (and also the case of allowing classes access to `async_lib` without stashing it in a `startup` method).

danielballan: It's worth thinking about in the long term whether we can keep most of this separation but make it possible to handle these issues. One possibility is using context vars to support a `what_is_my_server()`function. I would want to sit with that for awhile though; it feels a bit too fancy.

klauer:
Turns out 32-bit signed ints are not so great at displaying bytes of free memory; these will need to be in kilobytes (max 1TB RAM; should be 'nuff for anybody).

Added number of threads as an additional useful metric.

caproto/server/stats.py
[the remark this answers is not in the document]
klauer: Resolved in a8e3d7e

klauer:
Follow-up PR will have more tracemalloc filtering

Issue #647, opened 2020-09-03
Identify areas where unbounded dictionaries/lists/etc could be a problem


klauer:
Ran into a scenario where an IOC host consumed all of its memory. 

While it's unclear if caproto was the root cause of that memory consumption, it happened to be the machine running our first ever production caproto IOC (v0.6.0). It was interacting with likely hundreds of other CA clients indirectly through UDP.

```
[E 11:32:36.768       common:  768] Broadcaster command queue evaluation failed
    Traceback (most recent call last):
      File "(caproto-env)/caproto-dev/caproto/server/common.py", line 763, in broadcaster_queue_loop
        await self._broadcaster_queue_iteration(addr, commands)
      File "(caproto-env)/caproto-dev/caproto/server/common.py", line 819, in _broadcaster_queue_iteration
        self.broadcaster.process_commands(commands)
      File "(caproto-env)/caproto-dev/caproto/_broadcaster.py", line 155, in process_commands
        self._process_command(self.their_role, command, history)
      File "(caproto-env)/caproto-dev/caproto/_broadcaster.py", line 177, in _process_command
        self.unanswered_searches[command.cid] = command.name
    MemoryError
```

Unanswered searches here is technically unbounded, but its domain is limited to `(0...2**16 - 1)`, so it couldn't be the actual issue.

Regardless, there's a need here to do a deep dive and identify areas where unbounded items could bring down servers or even long-running clients.

tacaswell:
This is the broadcaster seeing searches from clients to hosts IOCs it is not sitting in front of so they never get cleared?

PR #650, opened 2020-09-04
API/FIX/CLN: do not track search requests on server; remove curio and trio clients

tacaswell:
The search related API was removed from :class:`ca.Broadcaster`, all
of the search request accounting is handled in the client code.  The
code that is used on the servers can not do this book keeping because
we can not know what other servers are out there and if the
SearchRequest actually got serviced (as that goes back uni-cast).

partially addresses #647

The tests I removed may need to be put back directly testing `ThreadsafeCounter`?


tacaswell:
The curio client used the base Broadcaster search logic :(

danielballan:
I would be OK with removing the experimental and very incomplete curio and trio clients until we get a chance to rebuild them in the fashion of our asyncio client.

tacaswell:
I think the fix is relatively straight forward (copy the search logic from asyncio -> curio?).  We probably should go over the search functionality in all of the clients because they are each a bit different at this point....

klauer:
Despite all of my efforts to keep curio/trio clients from breaking things, I'm on board with just nuking them for now.

klauer:
Removed and tests are passing. Did not look much into docs, though.

danielballan (approved):
The docs changes you made look great. Very clear. There were no curio or trio client docs to remove or update.

@tacaswell might be right that just fixing the curio and trio clients by copy/pasting the broadcaster logic might be _easier_ than excising those clients, but they are so far out of sync with the functionality and APIs of the other clients that I think removing them is the right thing to do. Thanks for doing the work.

tacaswell:
I did almost no work here...

PR #652, opened 2020-09-06
WIP: Log cache sizes in telemetry loop.

danielballan:
Just a start -- making public early to facilitate coordination with related efforts by @klauer, @tacaswell 

- [ ] Actually run the telemetry_loop in the respective server implementations
- [ ] Additional info beyond cache sizes

### DIFF
--- a/caproto/server/common.py
+++ b/caproto/server/common.py
@@ -1,16 +1,17 @@
-from collections import defaultdict, deque, namedtuple, ChainMap
-from collections.abc import Iterable
 import copy
 import logging
 import sys
 import time
 import weakref
-import caproto as ca
-from caproto import (apply_arr_filter, get_environment_variables,
-                     RemoteProtocolError, CaprotoKeyError, CaprotoRuntimeError,
-                     CaprotoNetworkError, ChannelType)
-from .._dbr import SubscriptionType, _LongStringChannelType
+from collections import ChainMap, defaultdict, deque, namedtuple
+from collections.abc import Iterable
 
+import caproto as ca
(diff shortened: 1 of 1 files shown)


danielballan:
Now my intent is to do this via PVs rather than logging, incorporated in the structure laid out by #649. Leaving open as documentation of the sketch until I get time to incorporate it.

danielballan:
@klauer I had intended to incorporate this in `StatsHelper` but I think it will run into the same limitation as [discussed here](https://github.com/caproto/caproto/pull/649#discussion_r494592502): it needs access to server state that we don't currently provide and may not _want_ to provide.

Given that should we:

1. Punt on this, deciding that the the tracemalloc stats are sufficient to debug runaway caches and that we don't want to take on an extra polling task and associated complexity in the code and logging output
2. Go ahead and implement this as originally suggestion
3. Open an issue to revisit this if we ever sort out how to cleanly let ChannelData / pvproperty access server state.

I am pretty ambivalent.

klauer:
I'm somewhat leaning toward (2), or maybe my tacked-on alternative (4) below.

(4) It'd be nice to have this exact information in some easy-to-access way. If/when we ever get the REPL-like interface going (or the originally proposed `SIGINFO`), we could then query the server state easily with a method such as what you've already written (sans the loop and threshold checking).

danielballan:
I like (4). Let’s push on that and then circle back here.

danielballan:
I'm going to close this because it's lower behind a lot of other things I'm more likely to work on for caproto.

PR #656, opened 2020-09-23
CI: cleanup

klauer:
`dataclasses` definitely is not necessary; repeater is started twice; unclear about the others. Let's see what happens.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -86,7 +86,7 @@ script:
   - rm -f `which caRepeater`
 
   - pip install git+https://github.com/klauer/catvs.git@py3k
-  # Work around pip bug that fails to respect trio's pin of attr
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -86,26 +86,19 @@ script:
   - rm -f `which caRepeater`
 
   - pip install git+https://github.com/klauer/catvs.git@py3k
-  # Work around pip bug that fails to respect trio's pin of attrs.
+  # Work around pip bug that fails to respect trio's pin of attr
```
klauer, line 89:
```suggestion
  # Work around pip bug that fails to respect trio's pin of attrs.
```

danielballan (approved):
:+1: Merge at will when CI passes

PR #660, opened 2020-09-25
CI: please, go faster

klauer:
I'm not sure if this should be merged or if I want it to be merged, even.

That being said, I'm very frustrated with the speed of caproto's CI and this is my way of complaining.

### DIFF
--- a/caproto/tests/test_cli_scripts.py
+++ b/caproto/tests/test_cli_scripts.py
@@ -1,11 +1,30 @@
-import subprocess
 import sys
+import uuid
 
 import pytest
 
+import caproto.commandline.get
+import caproto.commandline.monitor
+import caproto.commandline.put
 from caproto.sync.client import block, read, subscribe, write
 
-from .conftest import dump_process_output
+from .conftest import _caproto_ioc, _epics_base_ioc, dump_process_output
+
+
+@pytest.fixture(scope='module')
+def prefix():
+    'Random PV prefix for a server'
+    return str(uuid.uuid4())[:8] + ':'
+
+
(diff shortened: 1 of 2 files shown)


caproto/tests/test_cli_scripts.py
```diff
@@ -151,15 +191,15 @@ def test_subscribe_options(more_kwargs, ioc):
...
+def test_cli(command, args, ioc, monkeypatch):
     args = fix_arg_prefixes(ioc, args)
-    p = subprocess.Popen([sys.executable, '-um', 'caproto.tests.example_runner',
-                          '--script', command] + list(args),
-                         stdout=subprocess.PIPE, stderr=subprocess.PIPE
-                         )
-    _subprocess_communicate(p, command, timeout=10.0)
+    try:
+        call_script(monkeypatch, command, args)
+    except SystemExit as ex:
```
tacaswell, line 198: I did not know you could do this, but it makes sense...

tacaswell (approved):

tacaswell:
Have you tried adding `-n4` to the pytest invocation?

klauer:
I think CI is limited to 1 core?

tacaswell:
There are at least 2 on travis.  I also suspect that a lot of our time is network or disk not CPU bound so over-subscribing may be worth trying.

klauer:
I think parallelizing the CI with the changes in this PR would actually conflict (this shares the IOC)

tacaswell:
I would prefer if the tests stay isolated enough to run in parallel, that said pytest-xdist can be pretty smart about how it splits up (ex run a module serially).

PR #661, opened 2020-09-25
Reduce the minimum requirements to run tests

tacaswell:
Make the tests more forgiving of missing dependencies.

Without pyepics or epics_base you can still test that caproto can talk to its self.  Early on I think we needed the safety blanket of always testing against the reference, but at this point I think we can, at least locally, bootstrap the testing.


klauer:
Seems reasonable

PR #662, opened 2020-09-25
CI: parallelize?

klauer:
Just a test; nothing to see here

(will it actually pass?)


klauer:
Good call, @tacaswell - this gives the same gains as #660 without changing the tests.

tacaswell (approved):

PR #671, opened 2020-10-05
Reorganize ioc_examples.

### LINKED ISSUE #616: Examples are too clever
The example IOCs leave a lot to be desired - many were written to test the server functionality in place of a proper test suite.

These should be removed or cleaned and simplified.

danielballan:
Closes #616

The top-level directory contains only examples that are succinct and not overly
clever. The hope is too avoid bewildering new users by mixing straightforward
examples with pathological ones.

Summary of changes:

```
 rename caproto/ioc_examples/{ => hardwarwe_specific}/barcode_reader.py (95%)
 create mode 100644 caproto/ioc_examples/pathological/__init__.py
 rename caproto/ioc_examples/{ => pathological}/defaultdict_server.py (99%)
 rename caproto/ioc_examples/{ => pathological}/reading_counter.py (93%)
 rename caproto/ioc_examples/{ => pathological}/spoof_beamline.py (97%)
 create mode 100644 caproto/ioc_examples/too_clever/__init__.py
 rename caproto/ioc_examples/{ => too_clever}/areadetector_image.py (99%)
 rename caproto/ioc_examples/{ => too_clever}/caproto_to_ophyd.py (93%)
 rename caproto/ioc_examples/{ => too_clever}/caproto_to_typhon.py (93%)
 rename caproto/ioc_examples/{ => too_clever}/dynamic_pvgroups.py (96%)
 rename caproto/ioc_examples/{ => too_clever}/trigger_with_pc.py (97%)
 rename caproto/{ioc_examples => tests}/verify_with_catvs.py (87%)
```

As discussed in #616, examples are moved into `ioc_examples/too_clever/`,
`ioc_examples/pathological/`, or (in just one) into caproto's tests package.

I expect we won't be in 100% agreement about where things should go, but
hopefully this is a step in the right direction. I am happy for others to
push commits moving things around further.

I would only argue for `mini_beamline.py`. That one _is_ long but it's not
pathological or especially clever. Moreover, it is used by downstream things
(like tutorials) so moving it would have a larger impact that moving the others.

### DIFF
--- a/caproto/ioc_examples/hardware_specific/barcode_reader.py
+++ b/caproto/ioc_examples/hardware_specific/barcode_reader.py
@@ -4,8 +4,9 @@ This example requires a barcode reader and the Python library evdev.
 
 This only works on linux (as it uses the linux kernel input events).
 """
-from caproto.server import pvproperty, PVGroup, template_arg_parser, run
 from evdev import InputDevice, categorize
+
+from caproto.server import PVGroup, pvproperty, run, template_arg_parser
 
 
 class BarcodeIOC(PVGroup):
--- a/caproto/ioc_examples/mini_beamline.py
+++ b/caproto/ioc_examples/mini_beamline.py
@@ -1,4 +1,8 @@
 #!/usr/bin/env python3
+"""
+This example is quite large. It defines a range of simulated detectors and
+motors and is used for demos and tutorials.
+"""
 import contextvars
 import functools
 import math
--- a/caproto/ioc_examples/pathological/defaultdict_server.py
+++ b/caproto/ioc_examples/pathological/defaultdict_server.py
@@ -1,8 +1,9 @@
 #!/usr/bin/env python3
 
-from caproto.server import ioc_arg_parser, run
 from collections import defaultdict
+
 from caproto import ChannelData
+from caproto.server import ioc_arg_parser, run
 
 
 class ReallyDefaultDict(defaultdict):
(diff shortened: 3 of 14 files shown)


caproto/ioc_examples/hardwarwe_specific/barcode_reader.py
```diff
@@ -4,9 +4,10 @@
```
klauer, line 4: Typo in directory name

klauer (changes requested):
Overall, agree on where things were moved.

Azure CI needs updating due to new catvs IOC location.

klauer:
Afterthought - were any of the moved examples linked in the documentation?

danielballan:
Good question, but no.

danielballan:
> Azure CI needs updating due to new catvs IOC location.

Does 24ee451 fix this? I couldn't find any references to this in the Azure-specific files, and `git grep` only turned up the reference the 24ee451 fixes.

caproto/ioc_examples/hardwarwe_specific/barcode_reader.py
danielballan: Fixed, force pushed

klauer:
I think the fact that Azure failed much more quickly than Travis was what led to my comment above. Your fix is probably good - though looks like there's at least one more needing fixing, looking at the new failures.

klauer (approved):

klauer:
Random test failures seem to be back up again, meaning we should likely disable xdist parallel testing for the n-th time. 😦

klauer (approved):

PR #673, opened 2020-10-11
[PVA] ENH: initial scalar/scalararray normative type support

klauer:
(WIP, still playing around with this)

* `PVAGroup` with `attr = pvaproperty(value=0)` implies creating an attribute with `NTScalarInt64` (including the control struct, etc.)
* `PVAGroup` with `attr = pvaproperty(value=[0])` implies creating an attribute with `NTScalarArrayInt64` (including the control struct, etc.)


klauer:
On second thought, this is good enough for now. Will circle back later.

PR #675, opened 2020-10-15
CI: jobs

klauer:


klauer:
If we can find another source of epics-base binaries, we can remove epics-on-travis in this PR 🤞

klauer:
@danielballan @tacaswell objections to the change in the job matrix? If not, will squash/merge and try an alternate `softIoc` source the next chance I get to work on this.

klauer:
Tests passing -> no reply => implied agreement => _merge_

PR #679, opened 2020-10-17
CI: remove epics-on-travis, use conda + conda-forge epics-base

### LINKED ISSUE #676: CI: nuke epics-on-travis, rely on softIoc only
Protocol-wise, there's no difference between a motorSim IOC containing motor records and an epics-base `softIoc` with analog input/output records.

We have built-in support for per-test running of a background `softIoc` that doesn't require any previous setup (other than having epics-base available, of course), and we should use it.

This should increase reliability of CI jobs.

klauer:
Closes #676 

If I can get this working on travis, will change azure as well
Maybe even try GitHub CI for fun


klauer:
_The future of caproto CI is here_! I'm sure everyone's as excited as I am.

PR #681, opened 2020-10-30
FIX: be forgiving about dead Executors with dead circuits

tacaswell:
It seems during shutdown Python knocks oven the executor before we do
causing the notifications of channel disconnect to fail.

As this happens on process exit, just accept it and move on.

This empirically works, but I do not understand why as the code in the callbacks is in fact running before we get to the shutdown line!

### DIFF
--- a/caproto/threading/client.py
+++ b/caproto/threading/client.py
@@ -1440,7 +1440,18 @@ class VirtualCircuitManager:
                 event.set()
             callback = ioid_info.get('callback')
             if callback is not None:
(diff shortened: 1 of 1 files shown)


danielballan:
Looks good. We should fix the asyncio client in the same fashion while we are here.

tacaswell:
I am worried this is too forgiving.   

If make these raise you see protocol errors with trying to push the messages to stop updates through the state machine.

PR #683, opened 2020-11-17
ENH: add an escape hatch for pvproperty putters to skip writing (SkipWrite exception)

klauer:
Better ideas welcome


danielballan:
This seems like a good use for an exception to me.

Issue #691, opened 2020-12-16
Request: Warn/raise if setting both .scan and .startup


ronpandolfi:
I ran into an issue when trying to utilize both `.scan` and `.startup`; only one can be used at a time. This is noted in the docstring of `.scan`, but not `.startup`. It would be nice to give feedback when using these incorrectly (simultaneously).

danielballan:
Any interest in tackling this one @ronpandolfi. It seems it might be a "good first issue" for caproto server development/maintenance.

ronpandolfi:
@danielballan Do you mind if I raise ValueError in this situation?

danielballan:
Seems better than the status quo of silently failing for sure. Ideally we could compose them, but I’d need to think about what implications that might have. Happy to merge #691 as a step forward.

PR #692, opened 2020-12-16
Raise ValueError when assigning a startup if one already exists.

ronpandolfi:
In response to https://github.com/caproto/caproto/issues/691

### DIFF
--- a/caproto/server/server.py
+++ b/caproto/server/server.py
@@ -705,6 +705,10 @@ class pvproperty:
         """
         Usually used as a decorator, this sets ``startup`` in the PVSpec.
         """
+        if self.pvspec.startup:
+            raise ValueError(f'A startup function had already been assigned '
+                             f'to {self.pvspec}.')
+
         self.pvspec = self.pvspec._replace(startup=startup)
         return self
 


danielballan (approved):
Will merge in the next day or two unless @klauer has concerns.

danielballan:
Worth checking whether this breaks overriding startup in subclass. I don’t think it does but there be magick here....

klauer:
I think it might break easy subclassing and overriding of startup methods - that should be tested before proceeding.

danielballan (changes requested):
With fresh eyes, updating my review in accordance with my later comment and @klauer's.

klauer:
Need some more work on this one - it'll have to wait for the next minor tag.

I'm not sure how to differentiate the user's intent: "I want to override the startup method" versus "I accidentally assigned a startup method _and_ a scan method". Perhaps we could just support both `startup` and `scan`, since on the surface they don't seem like they should be exclusive to one another.

danielballan:
Yeah, I had the same thought.

klauer:
#729 intends to supersede this PR. Would it be OK to close this, @ronpandolfi?

klauer:
#729 is merged and hopefully an agreeable resolution. Closing.

Issue #680, opened 2020-10-27
windows error msg "OSError: [WinError 10057]" on _DatagramProtocol 


baldwinb:
Possibly related to https://caproto.github.io/caproto/v0.6.0/release-notes.html#v0-4-3-2020-01-29
using caproto 0.6.0
When running on windows with python 3.8.6 or 3.9.0 I see this error in the terminal every ~15s (but it still runs ok)
It runs without error on python 3.8.0 though
```[E 18:59:37.268        utils:   59] <caproto.asyncio.utils._DatagramProtocol object at 0x000002438CE269D0> receive error
    Traceback (most recent call last):
      File "C:\Program Files\Python39\lib\asyncio\proactor_events.py", line 523, in _loop_writing
        self._write_fut = self._loop._proactor.send(self._sock,
      File "C:\Program Files\Python39\lib\asyncio\windows_events.py", line 532, in send
        ov.WSASend(conn.fileno(), buf, flags)
    OSError: [WinError 10057] A request to send or receive data was disallowed because the socket
is not connected and (when sending on a datagram socket using a sendto call) no address was supplied```

**Comment by tacaswell** [comment]
That it is a datagram fail, this makes me think it is something in the search logic.   I wonder if we binding to interfaces we should not be for the broadcast?

**Comment by klauer** [comment]
I see this in `recvfrom` on Windows 10 with 3.8.6 with the new asyncio client, caproto v0.7.1:

```
[E 10:47:22.717        utils:   59] <caproto.asyncio.utils._DatagramProtocol object at 0x000001CD276FBAC0> receive error
    Traceback (most recent call last):
      File "C:\tools\miniconda3\envs\caproto\lib\asyncio\proactor_events.py", line 566, in _loop_reading
        self._read_fut = self._loop._proactor.recvfrom(self._sock,
      File "C:\tools\miniconda3\envs\caproto\lib\asyncio\windows_events.py", line 494, in recvfrom
        ov.WSARecvFrom(conn.fileno(), nbytes, flags)
    OSError: [WinError 10022] An invalid argument was supplied
```

Very simple script for reproduction:
```python
import asyncio
import sys

import caproto
from caproto.asyncio.client import Context


async def main(pvname):
    async with Context() as ctx:
        pv, = await ctx.get_pvs(pvname)
        data = await pv.read()
        print(pv, data)
        return val


if __name__ == "__main__":
    caproto.config_caproto_logging(level='DEBUG')
[...]

PR #707, opened 2021-01-25
FIX: win32 bind and shutdown errors in asyncio client

### LINKED ISSUE #708: asyncio client on Windows: per-Context "Event loop is closed" exception at teardown
Python 3.8+ (in which ProactorEventLoop supports the necessary `create_datagram_endpoint`)

Exception gets raised at teardown/exit after a context is created with `get_pvs` called:
```python
Exception ignored in: <function _ProactorBasePipeTransport.__del__ at 0x000002B0803E5DC0>
  File "C:\tools\miniconda3\envs\caproto\lib\asyncio\proactor_events.py", line 116, in __del__
    self.close()
  File "C:\tools\miniconda3\envs\caproto\lib\asyncio\proactor_events.py", line 108, in close
    self._loop.call_soon(self._call_connection_lost, None)
  File "C:\tools\miniconda3\envs\caproto\lib\asyncio\base_events.py", line 719, in call_soon
    self._check_closed()
  File "C:\tools\miniconda3\envs\caproto\lib\asyncio\base_events.py", line 508, in _check_closed
    raise RuntimeError('Event loop is closed')
RuntimeError: Event loop is closed
```

```python
import asyncio
import sys

import caproto
from caproto.asyncio.client import Context


async def main(*pvnames):
    async with Context() as ctx:
        for pvname in pvnames:
            pv = await ctx.get_pvs(pvname)
            data = await pv[0].read()
            print(pvname, data)


if __name__ == "__main__":
    # caproto.config_caproto_logging(level='DEBUG')
    asyncio.run(main(*sys.argv[1:]))
```

If more contexts are created, more exceptions are raised.

klauer:
Windows-related fixes. More users are trying the asyncio client on Windows and running into strange issues. Hopefully this will fix at least some of them.

Fixes
------
(1) There appears to have been a race for using the socket versus binding it, so `bind()` was moved up to avoid the following exception:
```
[E 11:09:50.604        utils:   61] <caproto.asyncio.utils._DatagramProtocol object at 0x000001F5E4BE8CD0> receive error
      File "C:\tools\miniconda3\envs\caproto\lib\asyncio\proactor_events.py", line 566, in _loop_reading
        self._read_fut = self._loop._proactor.recvfrom(self._sock,
      File "C:\tools\miniconda3\envs\caproto\lib\asyncio\windows_events.py", line 494, in recvfrom
        ov.WSARecvFrom(conn.fileno(), nbytes, flags)
```

(2) 
The proactor event loop handles shutdown and closing of sockets:

https://github.com/python/cpython/blob/7370be30017f81d2f41f1b4b2abf31dd9a3f8fb1/Lib/asyncio/proactor_events.py#L153-L168

So master currently duplicates that effort, causing asyncio's `finally` block to raise the following:
```
Exception in callback _ProactorBasePipeTransport._call_connection_lost(None)
handle: <Handle _ProactorBasePipeTransport._call_connection_lost(None)>
  File "C:\tools\miniconda3\envs\caproto\lib\asyncio\events.py", line 81, in _run
    self._context.run(self._callback, *self._args)
  File "C:\tools\miniconda3\envs\caproto\lib\asyncio\proactor_events.py", line 162, in _call_connection_lost
    self._sock.shutdown(socket.SHUT_RDWR)
OSError: [WinError 10038] An operation was attempted on something that is not a socket
```

I initially checked to see if this was the currently running event loop type, but then backed this up (in ac3d142) after running into (3) below, which I believe fixes it for both supported event loops.

(3) Event loop closed messages are avoided by trying to close the transport. This may be sufficient to get around (2) for all platforms, but I have not tested these separately.

```
Exception ignored in: <function _ProactorBasePipeTransport.__del__ at 0x000001E2DA11B310>
  File "C:\tools\miniconda3\envs\caproto\lib\asyncio\proactor_events.py", line 116, in __del__
    self.close()
  File "C:\tools\miniconda3\envs\caproto\lib\asyncio\proactor_events.py", line 108, in close
    self._loop.call_soon(self._call_connection_lost, None)
  File "C:\tools\miniconda3\envs\caproto\lib\asyncio\base_events.py", line 719, in call_soon
    self._check_closed()
  File "C:\tools\miniconda3\envs\caproto\lib\asyncio\base_events.py", line 508, in _check_closed
    raise RuntimeError('Event loop is closed')
RuntimeError: Event loop is closed
```

Note
-----
* https://docs.python.org/3.6/library/asyncio-eventloop.html#asyncio.AbstractEventLoop.create_datagram_endpoint
`create_datagram_endpoint` is _not_ supported until 3.8 on Windows, though it does _exist_. It may be worth making note of this in the documentation that people should be using 3.8+.
* `ProactorEventLoop` isn't importable on non-Windows platforms ~which made making a utility function for it sensible (in my opinion)~
 
Related issues
-------
#680 (may or may not close it)
Closes #708 

Test script
-----------

```python
import asyncio
import sys

import caproto
from caproto.asyncio.client import Context


async def main(*pvnames):
    async with Context() as ctx:
        for pvname in pvnames:
            pv = await ctx.get_pvs(pvname)
            data = await pv[0].read()
            print(pvname, data)


if __name__ == "__main__":
    # caproto.config_caproto_logging(level='DEBUG')
    asyncio.run(main(*sys.argv[1:]))
```

Using `softIoc` from conda-forge's epics-base on Windows.

### DIFF
--- a/caproto/asyncio/client.py
+++ b/caproto/asyncio/client.py
@@ -29,7 +29,7 @@ from ..client.search_results import (DuplicateSearchResponse, SearchResults,
                                      UnknownSearchResponse)
 from .utils import (AsyncioQueue, _CallbackExecutor, _DatagramProtocol,
                     _StreamProtocol, _TaskHandler, _TransportWrapper,
-                    get_running_loop)
+                    get_running_loop, is_proactor_event_loop)
 
 ch_logger = logging.getLogger('caproto.ch')
 search_logger = logging.getLogger('caproto.bcast.search')
@@ -1227,8 +1227,7 @@ class VirtualCircuitManager:
             sock.close()
 
         sock, self.socket = self.socket, None
-
-        if sock is not None:
+        if sock is not None and not is_proactor_event_loop():
             self._tasks.create(shutdown_old_socket(sock))
 
         tags = {'their_address': self.circuit.address}
--- a/caproto/asyncio/client.py
+++ b/caproto/asyncio/client.py
@@ -135,6 +135,12 @@ class SharedBroadcaster:
         await self._tasks.cancel_all(wait=True)
         self.log.debug('Broadcaster: Closing the UDP socket')
         self.udp_sock = None
+        try:
+            if self.protocol is not None:
+                if self.protocol.transport is not None:
+                    self.protocol.transport.close()
+        except OSError:
+            self.log.debug('Broadcaster transport close error')
 
         self.log.debug('Broadcaster disconnect complete')
 
@@ -1225,6 +1231,12 @@ class VirtualCircuitManager:
                 pass
 
             sock.close()
+
+        try:
+            if self.transport is not None:
+                self.transport.close()
+        except OSError:
+            self.log.debug('VirtualCircuitManager transport close error')
 
         sock, self.socket = self.socket, None
         if sock is not None and not is_proactor_event_loop():
--- a/caproto/asyncio/client.py
+++ b/caproto/asyncio/client.py
@@ -140,7 +140,7 @@ class SharedBroadcaster:
                 if self.protocol.transport is not None:
                     self.protocol.transport.close()
         except OSError:
-            self.log.debug('Broadcaster transport close error')
+            self.log.exception('Broadcaster transport close error')
 
         self.log.debug('Broadcaster disconnect complete')
 
--- a/caproto/asyncio/client.py
+++ b/caproto/asyncio/client.py
@@ -29,7 +29,7 @@ from ..client.search_results import (DuplicateSearchResponse, SearchResults,
                                      UnknownSearchResponse)
 from .utils import (AsyncioQueue, _CallbackExecutor, _DatagramProtocol,
                     _StreamProtocol, _TaskHandler, _TransportWrapper,
-                    get_running_loop, is_proactor_event_loop)
+                    get_running_loop)
 
 ch_logger = logging.getLogger('caproto.ch')
 search_logger = logging.getLogger('caproto.bcast.search')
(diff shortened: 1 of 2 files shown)


klauer:
The fix from (3) may be enough to fix (2) but I'm not 100% on it. Anecdotally, the test script works on OSX/Windows with this PR as of beff58b (and regardless of whether `shutdown_old_socket` is called on OSX).

klauer:
Any windows users up to giving this PR a try and reporting their results?

I suspect there is no one on our team actively working with Windows recently (or much of caproto at all, for that matter)

TheManderton:
Running this script on Windows 10, 2004 produces the error `OSError: [WinError 10022] An invalid argument was supplied`. According to stack overflow this is caused because the socket needs to be binded first? As a result, the search for PVs times out.

klauer:
Hmm, if you're actually running the branch from this PR, then the fix hasn't worked as intended. The message you noted is one this PR is trying to work around. (Can you double-check you're not running caproto master/0.7.1?)

You should be able to just do `python test_script.py pv_name1 pv_name2` (etc)

TheManderton:
My apologies for the late response, I think I must have done something wrong the first time, this is what I got now:

`>python3 poll.py 13SIM1:cam1:SizeX
13SIM1:cam1:SizeX ReadNotifyResponse(data=Array('i', [1024]), data_type=<ChannelType.LONG: 5>, data_count=1, status=CAStatusCode(name='ECA_NORMAL', code=0, code_with_severity=1, severity=<CASeverity.SUCCESS: 1>, success=1, defunct=False, description='Normal successful completion'), ioid=0, metadata=None)`

To me, that looks like what I think you were expecting?

klauer:
Yes, that looks good - no errors/exceptions other than that output, I hope? Thanks for giving it a go!

TheManderton:
Nope, that was the only thing printed on the console! So it looks like your fix worked. Thank you for considering us Windows users.

caproto/asyncio/client.py
```diff
@@ -135,6 +135,12 @@ async def disconnect(self):
         await self._tasks.cancel_all(wait=True)
         self.log.debug('Broadcaster: Closing the UDP socket')
         self.udp_sock = None
+        try:
+            if self.protocol is not None:
+                if self.protocol.transport is not None:
+                    self.protocol.transport.close()
+        except OSError:
+            self.log.debug('Broadcaster transport close error')
```
danielballan, line 143: If we make these `exception` instead of `debug` it will be easier to get detailed bug reports.

danielballan (approved):
One suggestion. Consider and merge at will.

caproto/asyncio/client.py
klauer:
```suggestion
            self.log.exception('Broadcaster transport close error')
```

PR #709, opened 2021-01-25
FIX: Win32 server beacon fix on proactor event loop

klauer:
May fix some beacon errors with running IOCs on Windows with the proactor event loop:

```
[E 12:47:49.911        utils:   59] <caproto.asyncio.utils._DatagramProtocol object at 0x0000011599F66820> receive error
      File "C:\tools\miniconda3\envs\caproto\lib\asyncio\proactor_events.py", line 563, in _loop_reading
        self._read_fut = self._loop._proactor.recv(self._sock,
      File "C:\tools\miniconda3\envs\caproto\lib\asyncio\windows_events.py", line 449, in recv
        ov.WSARecv(conn.fileno(), nbytes, flags)
[I 12:47:49.913       server:  206] Server startup complete.
```

Not fully tested, but anecdotally fixed the above

(Based on #707 , which should be merged first)

### DIFF
--- a/caproto/asyncio/client.py
+++ b/caproto/asyncio/client.py
@@ -29,7 +29,7 @@ from ..client.search_results import (DuplicateSearchResponse, SearchResults,
                                      UnknownSearchResponse)
 from .utils import (AsyncioQueue, _CallbackExecutor, _DatagramProtocol,
                     _StreamProtocol, _TaskHandler, _TransportWrapper,
-                    get_running_loop)
+                    get_running_loop, is_proactor_event_loop)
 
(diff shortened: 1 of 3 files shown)


klauer:
Looks like I messed up the PR a bit, bringing in some unrelated changes. The core fix in this PR is important to get merged for Windows users on the proactor event loop.

(#720 may change how this works, so additional windows testing should be done then)

klauer:
Scratch that. I can't even figure out where the fix was here.
Will have to revisit and re-test on windows after #720 .

PR #723, opened 2021-02-17
Add clear expectation-setting language.

danielballan:
Questions about whether it makes sense to "bet" on caproto and invest in caproto-based infrastructure have come from multiple facilities in the past couple months. It makes us happy to learn about caproto adoption, but we worry about caproto become _too_ popular to ~2 core devs to support mostly in our evenings/weekends.


klauer (approved):
Well-put, as usual! 👍

PR #727, opened 2021-03-02
FIX: CAPROTO_DEFAULT_TIMEOUT interpreted as string

### LINKED ISSUE #726: CAPROTO_DEFAULT_TIMEOUT should be "float" value
The current implementation (`caproto 0.7.1`) reads the environment variable `CAPROTO_DEFAULT_TIMEOUT` which implicitly has the type `string`. This causes calculations of connection timeouts to fail, such as `deadline = time.monotonic() + timeout`, because the interpreter is trying to add types "float" + "str".

File "[caproto/caproto/threading/client.py](https://github.com/caproto/caproto/blob/master/caproto/threading/client.py)":

```python
[line 49]    GLOBAL_DEFAULT_TIMEOUT = os.environ.get("CAPROTO_DEFAULT_TIMEOUT", 2)
```

Suggest to cast this value to "float"...

```python
[line 49]    GLOBAL_DEFAULT_TIMEOUT = float(os.environ.get("CAPROTO_DEFAULT_TIMEOUT", 2))
```

klauer:
Closes #726 

Took this as an opportunity to slightly rework the threading client and sync client to use the "common" client timeouts and sentinel objects. We had pulled these out when starting the asyncio client, if I recall correctly.

Shared settings accessed by way of `caproto.client.common.xyz` can more readily be changed/tweaked than if we import the constants directly (in certain use cases, at least, not when used as function arguments), leading to the style proposed in the client changes.


klauer:
Looks like you had done some work on this a couple years ago, @mrakitin - have any time to look through this as @danielballan has been so swamped?

mrakitin (approved):
@klauer, I am definitely in favor of the central configuration, as it's implemented here. I see no issues with the updates.

PR #729, opened 2021-03-02
REF: allow for both 'scan' and 'startup' hooks in server

### LINKED ISSUE #691: Request: Warn/raise if setting both .scan and .startup

klauer:
Supersedes #692 
Closes #691 (by allowing both, rather than raising)

Tested locally and now working.


danielballan (approved):
Looks right. Would be nice to test start, scan, and both to be sure.

PR #730, opened 2021-03-03
ENH: partially support RVAL for mbbi, mbbo, bi, and bo records

### LINKED ISSUE #724: Add support for RVAL fields
Currently support is missing for RVAL fields. I ran into this when trying to use an mbbi mock record.

klauer:
Closes #724 

Record generation update PR: https://github.com/caproto/reference-dbd/pull/1/files

Supports read-only access of RVAL for mbbi, mbbo, bi, and bo records.

For the future, the field instance needs a hook to see when the pvproperty (i.e., ".VAL") has updated so writes and other things can be supported more easily. I don't think that hook will be too difficult, but it will require some work.

### DIFF
--- a/caproto/_data.py
+++ b/caproto/_data.py
@@ -759,6 +759,14 @@ class ChannelEnum(ChannelData):
 
     enum_strings = _read_only_property('enum_strings')
 
+    @property
+    def raw_value(self) -> int:
+        """The raw integer value index of the enum string."""
+        try:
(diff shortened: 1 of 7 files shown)


danielballan:
Maybe @baldwinb would be a good reviewer for this?

klauer:
I'm planning on more sweeping changes for records. Closing this for now to avoid conflicts.

PR #731, opened 2021-03-03
FIX/ENH: attempt to support array interface to remove bottleneck processing numpy arrays

### LINKED ISSUE #705: Preprocessing value bottlenecks
Background
------------

Preprocessing conversions of data happen in `ChannelData` to keep values passed along by any source - in many different potential forms - stored in a consistent manner:
https://github.com/caproto/caproto/blob/03cff19954c81a40e679d592d26b6a9277032b65/caproto/_data.py#L265-L298

These conversions can be tricky and break things unexpectedly. Changes should be approached cautiously - with more test cases added where missing/needed.

Problem
----------

There is special preprocessing for byte strings:
https://github.com/caproto/caproto/blob/03cff19954c81a40e679d592d26b6a9277032b65/caproto/_data.py#L980-L1006

And strings with associated encodings:
https://github.com/caproto/caproto/blob/03cff19954c81a40e679d592d26b6a9277032b65/caproto/_data.py#L1036-L1060

Particularly the `map()` call in these can be an awful bottleneck - especially when the passed-in value starts of as either an already-valid `array.array` or a `np.ndarray` (as shown to perform terribly in #704). 

Possibility 1
---------------
Backends could perform this processing in a way that makes sense for them - numpy can handle its `ndarray`, and `array` can handle these string/byte conversions appropriately - and quickly.

Possibility 2
-------------
* Create some string/byte-handling utilities outside of the backends
* Possibly use the array interface (https://numpy.org/doc/stable/reference/arrays.interface.html) to not require numpy here as well
* Alternatively, a `try/except ImportError` for `numpy`

Other ideas
------------
Any other ideas/thoughts?

klauer:
Closes #705 

Overview
----------
Supports the numpy array interface to remove a significant ``O(n)``  bottleneck when pre-processing numpy arrays for ``ChannelByte`` instances.  

Notes
------
* To avoid an additional copy on immutable data, set ``value.flags.writeable``  to ``False`` when using compatible numpy arrays.  (#731)
* ``ChannelChar`` will eventually be decoded, so `.tobytes()` is unavoidable - but we can at least let numpy take care of this for us instead of using `map(chr, value)`.

### DIFF
--- a/caproto/_data.py
+++ b/caproto/_data.py
@@ -735,25 +735,34 @@ class ChannelData:
         return AccessRights.READ | AccessRights.WRITE
 
     def is_compatible_array(self, value) -> bool:
-        try:
-            interface = value.__array_interface__
-        except AttributeError:
+        """
+        Check if the provided value is a compatible array.
+
+        This requires that ``value`` follow the "array interface", as defined by
+        `numpy <https://numpy.org/doc/stable/reference/arrays.interface.html>`_.
+
+        Parameters
+        ----------
+        value : any
+            The value to check.
+
+        Returns
+        -------
+        bool
+            True if ``value`` is compatible, False otherwise.
+        """
+        interface = getattr(value, "__array_interface__", None)
+        if interface is None:
             return False
 
-        try:
-            dimensions = len(interface['shape'])
-            strides = interface['strides']
-            typestr = interface['typestr']
-        except KeyError:
-            return False
-
+        dimensions = len(interface['shape'])
         return (
             # Ensure it's 1 dimensional:
             dimensions == 1 and
             # Not strided - which 1D data shouldn't be anyway...
-            strides is None and
+            interface['strides'] is None and
             # And a compatible array type, defined in the class body:
-            typestr in self._compatible_array_types  # compatible type
+            interface['typestr'] in self._compatible_array_types
         )
 
 
--- a/caproto/_data.py
+++ b/caproto/_data.py
@@ -17,7 +17,8 @@ from ._dbr import (DBR_STSACK_STRING, DBR_TYPES, AccessRights, AlarmSeverity,
                    SubscriptionType, _channel_type_by_name,
                    _LongStringChannelType, native_type, native_types,
                    time_types, timestamp_to_epics)
-from ._utils import CaprotoError, CaprotoValueError, ConversionDirection
+from ._utils import (CaprotoError, CaprotoValueError, ConversionDirection,
+                     is_array_read_only)
 
 __all__ = ('Forbidden',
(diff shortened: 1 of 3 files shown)


caproto/_data.py
```diff
@@ -1041,6 +1068,8 @@ def preprocess_value(self, value):
                 value = b''
             elif len(value) == 1:
                 value = value[0]
+            elif self.is_compatible_array(value):
+                value = value.tobytes()
```
danielballan, line 1072: This costs a memory copy. I wonder if memoryview would work.

klauer: Goal of `preprocess_value` as I recall is to make it a consistent `bytes` string for the `putter` handler to accept/reject/modify. 

I suppose if the client code is _writing_ such arrays directly, then it may be assumed that their handlers can accept such arrays and won't presume a `bytes`-like interface. I'm not sure there's a 100% correct answer here.

The alternative would be just allowing the array through unchanged. Storing such a mutable value could lead to problems unless we copy it down the line (another thing I'd have to double-check...).

caproto/_data.py
```diff
@@ -733,6 +734,28 @@ def check_access(self, hostname, username):
         """
         return AccessRights.READ | AccessRights.WRITE
 
+    def is_compatible_array(self, value) -> bool:
+        try:
```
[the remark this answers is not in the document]
klauer: I'd assume `getattr(obj, attr, nonexistent_default_sentinel)` would be preferable. `hasattr` then `getattr` is race condition-prone, even if it's really uncommon - except perhaps when tearing down an object.

What is your preferred `hasattr`-based snippet? Is the possibility of `delattr` between accessing something that's just an acceptable possibility?

caproto/_data.py
```diff
@@ -733,6 +734,28 @@ def check_access(self, hostname, username):
...
+            interface = value.__array_interface__
+        except AttributeError:
+            return False
+
+        try:
+            dimensions = len(interface['shape'])
+            strides = interface['strides']
+            typestr = interface['typestr']
+        except KeyError:
+            return False
```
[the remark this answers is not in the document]
klauer: 🤷‍♂️

caproto/_data.py
klauer: I swayed your opinion, and now I'll try to sway it back.

`.tobytes()` ends up being too slow for large arrays, but we _can_ copy it and still maintain a reasonably fast update speed.
The array interface provides an option to check if the array is flagged as read-only. The read-only nature is only a matter of user-intent and isn't _actually_ read-only, of course. 

How do you feel about the recently pushed commits?

PR #734, opened 2021-03-09
REF: rework auto-generation of records (again) and restructure

### LINKED ISSUE #724: Add support for RVAL fields

klauer:
Pairs with https://github.com/caproto/reference-dbd/pull/2

This is largely just shuffling code around in the interest of making field support easier to write and maintain.
That said, I think this is an important step forward for _better_ record field support. 

Main points:
* `caproto.server.records` is now a module (with back-compat attributes)
* `caproto.server.records.base` is auto-generated, along with `.menus`
* `caproto.server.records.records` is in place for us to modify
    * Functionality previously (confusingly) shoved into jinja templates is now in user-modifiable code in this module
* Better organization with mixins, utilities (registry) in separate submodules

Additional functionality:
* A `value_write_hook` is now supported when updating a value in records.
    * This means that if the user updates `.VAL`, other fields will have the chance to update their state
    * An example which closes #724 is provided (`.VAL` updates `.RVAL` and vice-versa for bi/bo/mbbi/mbbo records)

Fixes:
* Inheritance by default shares `pvproperty` instances and therefore their custom putter/etc hooks
    * -> Shallow-copy now performed for record fields to avoid this.
    * Should revisit this with the broader scope of `PVGroup` in general as I'd classify this as a bug


klauer:
Tests passing; users reticent; PR merged

PR #736, opened 2021-03-09
REF: rework PVSpec instantiation, allowing for use outside of PVGroups

klauer:
Background
==========
Users have commented that `PVGroup` is _convoluted and confusing_ (@mattgibbs and others).

While I don't disagree about the convoluted nature, I think the cleanliness of the resulting `PVGroup` classes should be reason enough to forgive the implementation details.  

For those that feel that way, the alternative at the moment is to use `ChannelData` directly. This is fairly barebones and doesn't support any of the nice hooks that `pvproperty` brought along with it.

`PVSpec` was initially written with the idea that `pvproperty`/`PVGroup` could be thrown away and the useful bits could be retained. That work had never been completed until now.

This PR
=======
This PR allows you to use `PVSpec` on its own in a sort of "expert" mode, but still having the putter/startup/scan/record hooks that `PVGroup` brings with it.

It removes a significant portion of the convoluted instantiation that confuses users, and I think provides a reasonable middle-ground between "use bare ChannelData and perhaps reinvent the wheel" and "use `PVGroup` that you can't understand/despise/etc"

Example
=======
What does this look like?

Single PV + record example, partial of the one included in the PR:
```python
#!/usr/bin/env python3
from caproto.server import PVSpec, ioc_arg_parser, run


async def put_handler(instance, value):
    print(f"You wrote {value} to {instance.pvname}.")


pv_a = PVSpec(
    name="a",
    value=1,
    record='longin',
    doc='An integer',
    put=put_handler,
)

if __name__ == '__main__':
        default_prefix='pvspec:',
        desc="A simple IOC without using the PVGroup-style class syntax."
    )

    # It's up to you to add on the prefix and support any other ioc_options:
    pvdb = {
        ioc_options['prefix'] + 'a': pv_a.create()
    }
    run(pvdb, **run_options)
```

### DIFF
--- a/caproto/ioc_examples/decay.py
+++ b/caproto/ioc_examples/decay.py
@@ -1,8 +1,10 @@
 #!/usr/bin/env python3
-from caproto.server import pvproperty, PVGroup, ioc_arg_parser, run
-import numpy as np
 import time
 from textwrap import dedent
+
+import numpy as np
+
+from caproto.server import PVGroup, ioc_arg_parser, pvproperty, run
 
 
 class Decay(PVGroup):
@@ -38,10 +40,10 @@ class Decay(PVGroup):
 
     readback = pvproperty(value=0, dtype=float, read_only=True,
                           name='I',
-                          mock_record='ai')
+                          record='ai')
 
     done = pvproperty(value=0, dtype=float, read_only=True,
-                      mock_record='ai')
+                      record='ai')
 
     setpoint = pvproperty(value=100, dtype=float, name='SP')
     K = pvproperty(value=10, dtype=float)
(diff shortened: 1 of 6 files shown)


klauer:
New signature checker makes it easier to locate where your problem is:
```
RuntimeError: B (/Users/klauer/Repos/caproto/caproto/ioc_examples/records.py:25) must be callable with a signature like the
following:
    async def B(group, instance, value)
```

danielballan:
I really like that this provides a way to run a useful IOC without subclassing anything. It's nice that `ioc_arg_parser` works the same as usual. I think it helps connect the dots between the `pvdb` dict-of-`ChannelData` examples and the more common `PVGroup` examples.

PR #749, opened 2021-03-13
CI: add basic GitHub actions for Windows and macOS testing

### LINKED ISSUE #745: Windows: asyncio server "No service is operating at the destination network endpoint on the remote system"
This one on the server seems to interrupt further search replies:

```python
[E 13:50:27.545       server:  230] Broadcaster failed to receive on 0.0.0.0
      File "C:\mambaforge\envs\py38\lib\asyncio\proactor_events.py", line 548, in _loop_reading
        res = fut.result()
      File "C:\mambaforge\envs\py38\lib\asyncio\windows_events.py", line 812, in _poll
        value = callback(transferred, key, ov)
      File "C:\mambaforge\envs\py38\lib\asyncio\windows_events.py", line 500, in finish_recv
        return ov.getresult()
    OSError: [WinError 1234] No service is operating at the destination network endpoint on the remote system
```

### LINKED ISSUE #746: Windows: sync client ConnectionResetError
Per https://docs.microsoft.com/en-us/windows/win32/api/winsock/nf-winsock-recvfrom
> WSAECONNRESET
> The virtual circuit was reset by the remote side executing a hard or abortive close. The application should close the socket; it is no longer usable. On a UDP-datagram socket this error indicates a previous send operation resulted in an ICMP Port Unreachable message. 

The sync client sometimes fails to search with the above error during `recvfrom`.

### LINKED ISSUE #747: Clients: User-requested `context.disconnect()` attempts reconnection
Fix should be something along the lines of:
`self._disconnect(reconnect=False)`

This is also where the thread pool executor needs to be cleaned up.

### LINKED ISSUE #748: getsockname() called before bind in curio, trio servers
Resolution: call it after, or win32 will squawk

klauer:
Notably these do not rely on `conda` and do **not** test with their respective platform's `softIoc` executable like the azure pipelines tests do for Linux. So if they pass, at least caproto itself is internally consistent, but untested with respect to epics-base servers.

Also includes some fixes found along the way:

Closes #747 
Closes #746 
Closes #745 
Closes #748

Long-standing awful test "test_multithreaded_many_write" has been improved here, too. (As much as I like the concept behind that test suite, it's pretty poorly written...)

(Apologies for starting Actions as a push to master; I'm new to it and not sure if there was a way around it. Regardless, this PR should give a good initial first test suite runner)

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -1,4 +1,4 @@
-name: Python package
+name: caproto-actions
 
 on: [push, pull_request]
 
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -1,6 +1,6 @@
 name: Python package
 
-on: [push]
+on: [push, pull_request]
 
 jobs:
   build:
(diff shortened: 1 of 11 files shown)


klauer:
1 catvs test failure in 3.8-mac, 1 in 3.6-win.

Of 12 different test suites run, that's not so bad.

PR #750, opened 2021-03-13
FIX: call correct disconnect() method

klauer:
I'll get this one right eventually. Let's see how this does.


klauer:
14 successful on the first try! Certainly a fluke

PR #751, opened 2021-03-15
API: remove connection_state_callback, access_rights_callback from thread PV init

### LINKED ISSUE #609: Revisiting caproto.threading.client.PV.__init__
Tangential to #607:

I'm not sure why `PV.__init__`, which is internal, needs to take in `connection_state_callback` and `access_right_callback`. The public methods `PV.connection_callback.add_callback` and `PV.access_rights_callback.add_callback` can be called after `__init__` time, and will still be run even if registration happens after the relevant event (e.g when connection happens). And unlike `__init__`, they expose the possibility of registering multiple callbacks.

Would anyone be opposed to removing those parameters from `PV.__init__`? The public way to initialize `PV` is via `Context.get_pvs` and `PV.__init__` has always been documented as internal, not to be called by the user.

klauer:
Internal API, but API nonetheless

Closes #609

### DIFF
--- a/caproto/threading/client.py
+++ b/caproto/threading/client.py
@@ -1029,7 +1029,7 @@ class Context:
                 try:
                     pv = self.pvs[(name, priority)]
                 except KeyError:
-                    pv = PV(name, priority, self, None, None, timeout)
+                    pv = PV(name, priority, self, timeout)
                     names_to_search.append(name)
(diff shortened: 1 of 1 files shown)


klauer:
1 random failure among 13000 or so, not bad.

PR #771, opened 2021-05-27
ENH/REF: Support raw timestamps in ChannelData

### LINKED ISSUE #770: Support raw timestamps in ChannelData (no float conversion for beam pulse ID/fiducials, etc)
Current supported API:
```python
# Current method, passing in as a time.time() timestamp:
await data.write(value, timestamp=time.time())
```

Possible API additions could be as follows:
```python
# Pass timestamp in as a tuple:
await data.write(value, timestamp=(seconds_int, nanoseconds_int))
# Or using the already available data structure:
await data.write(value, timestamp=caproto.TimeStamp(seconds, nanoseconds))
```

Fortunately, timestamps are hidden away in ChannelData (`ChannelData._data["timestamp"]`) and only accessed by way of properties. The encapsulation should hopefully make it a bit easier to support either method.

Minimal modifications required:
* `ChannelData.timestamp` read-only property
* `ChannelData.epics_timestamp` read-only property
* Some `caproto._dbr` fixes which may be more obvious when I poke around with it

Clients, I believe, have easy access to the raw timestamp, but that should be double-checked.

cc @ddamiani

klauer:
Closes #770 

Hopefully doesn't break anything. `ChannelData._data[*]` is not intended to be modified by the user, and this PR uses that to its advantage.

I don't particularly like the class `TimeStamp`, but it _is_ already accessible to the user. Added a few convenience methods to it accordingly.

Also - more docstrings and improved ones, based on the `timestamp` change.

### DIFF
--- a/caproto/_data.py
+++ b/caproto/_data.py
@@ -4,6 +4,7 @@
 # data as a certain type, and they push updates into queues registered by a
 # higher-level server.
 import copy
+import datetime
 import time
 import weakref
 from collections import defaultdict, namedtuple
@@ -14,9 +15,9 @@ from ._commands import parse_metadata
 from ._constants import MAX_ENUM_STATES, MAX_ENUM_STRING_SIZE
 from ._dbr import (DBR_STSACK_STRING, DBR_TYPES, AccessRights, AlarmSeverity,
                    AlarmStatus, ChannelType, GraphicControlBase,
-                   SubscriptionType, _channel_type_by_name,
+                   SubscriptionType, TimeStamp, _channel_type_by_name,
                    _LongStringChannelType, native_type, native_types,
-                   time_types, timestamp_to_epics)
+                   time_types)
 from ._utils import (CaprotoError, CaprotoValueError, ConversionDirection,
                      is_array_read_only)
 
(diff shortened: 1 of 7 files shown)


caproto/_data.py
```diff
@@ -281,8 +283,13 @@ def __init__(self, *, alarm=None, value=None, timestamp=None,
...
 
         value = self.preprocess_value(value)
-        self._data = dict(value=value,
-                          timestamp=timestamp)
+
+        # The following _data isn't meant to be modified directly. It should be
+        # modified by way of the ``write()`` and ``write_metadata`` methods.
+        self._data = dict(
+            value=value,
+            timestamp=TimeStamp.from_flexible_value(timestamp),
```
danielballan, line 291: I like this use of a classmethod constructor.

caproto/_data.py
```diff
@@ -869,18 +909,25 @@ async def write_metadata(self, publish=True, units=None, precision=None,
...
-        'EPICS timestamp as (seconds, nanoseconds) since EPICS epoch'
-        return timestamp_to_epics(self._data['timestamp'])
+    def timestamp(self) -> float:
+        """UNIX timestamp in seconds."""
+        return self._data["timestamp"].timestamp
+
+    @property
+    def epics_timestamp(self) -> TimeStamp:
+        """EPICS timestamp as (seconds, nanoseconds) since EPICS epoch."""
+        return copy.copy(self._data["timestamp"])
```
danielballan, line 930: Part of me wants to suggest that it's fine to return the original here --- if you modify that struct you deserve what happens to you --- but you are probably right about this.

danielballan (approved):

danielballan:
One musing comment, no suggested changes. Merge at will!

caproto/_data.py
klauer: I'm 50/50 on it as well, and not to mention lazy. So I think the current implementation wins out!

PR #772, opened 2021-06-03
DOC: release notes for v0.8.1

klauer:
* pyepics compat fix
* Raw timestamp addition

Today is release day 🎆


klauer:
Tests passing; features previously approved. Let's go.

Issue #674, opened 2020-10-13
pvAccess support


klauer:
What's on my path forward for pvAccess support? Here's a brain dump which I'll try to keep up-to-date.

Overall / Core
--------------
- [x] All required messages implemented for get, put, monitor, RPC
- [x] Base clients/servers around dataclasses
- [x] Documentation outline + getting started (with sufficient warning given nature of pva support)
- [ ] Consider replacing (optional) parsimonious requirement with (optional) lark requirement
- [ ] Review/change what gets imported to the pva "top level" (`caproto.pva.*`)

Client
--------
- [x] Add command line "basic" tools: get, put, monitor
- [ ] Solidify command-line options. Re-add convenient scripts such as ``caproto-pva-get`` (or something less unwieldy).
- [ ] Add command line "extra" tools: pvlist, rpc/call
- [ ] Add asyncio client

User-facing Server API
--------
- [x] Working asyncio server (akin to dictionary-based servers in caproto V3)
- [x] Working PVAGroup + pvaproperty-based server
- [x] pvaproperty startup, shutdown, getter, putter, rpc
- [x] Preliminary normative type support (scalar, array; subclass checking)
- [x] Rename caproto.pva.ioc_examples.basic -> `custom_advanced` or something
- [ ] Further normative type support
- [ ] Consider removing `pvaproperty.get`
- [ ] Better RPC interface, like `caproto.server.pvfunction`

Server core
-------------
- [ ] Finish monitor pipelining support
- [ ] Allow access of data over Channel Access as well as pvAccess
- [ ] Update supported timestamps on write (where appropriate; record processing in V4 is still confusing to me); guaranteed to work for normative types

Testing
--------
- [ ] More/better testing
- [ ] Search out complex examples to verify serialization routine correctness

klauer:
See #773

PR #774, opened 2021-08-04
REF: remove pvAccess support

### LINKED ISSUE #773: Remove pvAccess support
* At over a year since I introduced it into caproto, there have been 0 usage attempts of its pvAccess support
* I haven't used it for anything myself
* I don't have the bandwidth to support it now, nor will I if someone comes along to use it

Avoiding the sunk cost fallacy (perhaps a few hundred hours of development effort?), I plan to just remove it entirely from the code base.

Well, at least I learned something along the way.

klauer:
Closes #773
See also #674

This is likely the first PVA-related PR @tacaswell will be delighted to review!


klauer:
Docs failures due to sphinx upgrade (I think); random test failures on GH actions, but all passing on azure.

Perhaps now that NSLS-II has agreed to maintain caproto, one of the first tasks could be to fix the reasons behind the documentation building failures?

PR #776, opened 2021-10-29
Fix versioneer compat with py311

tacaswell:


tacaswell:
Self-merging to as prep for a release.

PR #779, opened 2021-11-17
ENH: provide defaults for user limits

clintonroy:
Make it easy to set user limits, to mock out more of the EPICS database.


klauer (approved):
Thanks for the contribution, @clintonroy! I think this looks good.

(Oh, and don't mind our CI test failures below; some are at least passing. Someone has some maintenance work to do on the CI pipelines...)

Issue #783, opened 2022-03-22
BUG: Using lower-level PVSpec with kwarg `read_only=True` results in a PV that is not read only.


marrable:
I am relatively new to caproto, so I apologise if this is the intended behaviour and I am missing the necessary context.

#### Behaviour:
I was writing a simple IOC based on the [`no_pvproperty.py`](https://github.com/caproto/caproto/blob/master/caproto/ioc_examples/no_pvproperty.py) example and realised that even when I had specified `read_only=True` for a boolean value PVSpec, I was still able to change the PV at runtime via `caput <PV_NAME> <VALUE>` on a separate terminal.

#### Steps to Reproduce:
``` Python
#!/usr/bin/env python3
from caproto.server import PVSpec, ioc_arg_parser, run

pv_a = PVSpec(
    name="a",
    value=True,
    record='bi',
    doc='A bool',
    read_only=True
)

pvdb = {
    pvspec.name: pvspec.create(group=None)
    for pvspec in [pv_a]
}

def main():
    global pvdb, pv_a
    print("pvdb contents:", pvdb)
    print(pv_a.read_only)

    ioc_options, run_options = ioc_arg_parser(
        default_prefix='pvspec:',
        desc="A simple IOC without using the PVGroup-style class syntax."
    )
    # It's up to you to add on the prefix and support any other ioc_options:
    pvdb = {
        ioc_options['prefix'] + name: data
        for name, data in pvdb.items()
    }
    run(pvdb, **run_options)
```
Running this sample above (which is a stripped back version of the `no_pvproperty.py` example) perform a `caput` in a separate terminal to change the PV.

#### Potential Root Cause
In [`server.py`](https://github.com/caproto/caproto/blob/master/caproto/server/server.py) the following type maps are defined:
``` Python
pvspec_type_map = {
    str: PvpropertyChar,
    bytes: PvpropertyByte,
    int: PvpropertyInteger,
    float: PvpropertyDouble,
    bool: PvpropertyBoolEnum,
    enum.IntEnum: PvpropertyEnum,

    ChannelType.STRING: PvpropertyString,
    ChannelType.INT: PvpropertyShort,
    ChannelType.LONG: PvpropertyInteger,
    ChannelType.DOUBLE: PvpropertyDouble,
    ChannelType.FLOAT: PvpropertyFloat,
    ChannelType.ENUM: PvpropertyEnum,
    ChannelType.CHAR: PvpropertyChar,
}

# Auto-generate the read-only class specification:
[...]

PR #784, opened 2022-03-22
Updated method get_data_class() in PVSpec so that type maps are correct for read only

marrable:
Fix for issue described in: https://github.com/caproto/caproto/issues/783 .


tacaswell (approved):

klauer:
Thanks, @marrable!

PR #787, opened 2022-05-01
CI: add py310 and drop py36

tacaswell:


tacaswell:
Linux tests pass up to py310 and all but 1 osx 🎉

caproto/asyncio/client.py
```diff
@@ -1071,18 +1065,9 @@ async def send(self, *commands, extra=None):
...
-        async def _socket_send(buffers_to_send):
-            'Send a list of buffers over the socket'
-            try:
-                return self.transport.sock.sendmsg(buffers_to_send)
-            except BlockingIOError:
-                raise ca.SendAllRetry()
-
-        async with self._raw_lock:
-            await ca.async_send_all(buffers_to_send, _socket_send)
+        # TODO: can we trust asyncio to get this all out the door 😱
```
klauer, line 1068: Not sure we can. I seem to recall issues with large arrays back in the day.

This is a pretty major change though for the client, also requiring that the joined buffer be created rather than using the buffers by reference as created in `circuit.send`.

caproto/asyncio/utils.py
```diff
@@ -67,22 +67,20 @@ class _TransportWrapper:
 
     Parameters
     ----------
-    sock : socket.socket
-        The socket.
+    reader : ??
```
klauer, line 70: I think these were the stream ones: https://docs.python.org/3/library/asyncio-stream.html#asyncio.StreamReader

caproto/asyncio/client.py
tacaswell: fair.

After I wrote this I did a bit more reading and I think there is a better option that does not require the joining.

klauer:
They weren't intended to fail - especially not on just Windows. At what commit did they start failing?

PR #788, opened 2022-05-04
Use RotatingFileManager when restoring autosave

mfurseman:
If a `RotatingFileManager` is provided when setting up an `AutosaveHelper` it attempts to use the default filename, `autosave.json`, instead of the filename provided by the `RotatingFileManager` when restoring the autosaved values. This pull request changes that behaviour so the AutoSaveHelper instead loads the filename as specified within its `RotatingFileManager`.


klauer (approved):
Looks good to me - thanks, @mfurseman

PR #792, opened 2022-06-02
WIP/ENH: some annotations for PVGroups and SubGroups

klauer:
## Changes

* Annotations for the `pvproperty`, `SubGroup` descriptors' `__get__` methods
* Stand-in class for `get_read_write_pair` so that `.readback` and `.setpoint` show up with type hints. It works but it's not quite accurate

## Notes

* Would be nice to get generics working correctly with the `SubGroup` descriptor (`SubGroup[ThePVGroup](ThePVGroup, ...)` could then allow for digging into `ThePVGroup` attrs), perhaps even removing the need for the first argument altogether
* Suggestions welcome as this is at the limits of my type hinting abilities

### DIFF
--- a/caproto/server/server.py
+++ b/caproto/server/server.py
@@ -6,6 +6,8 @@ a single asyncio library.
 
 For an example server implementation, see caproto.curio.server
 """
+from __future__ import annotations
+
 import argparse
 import copy
 import enum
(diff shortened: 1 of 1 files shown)


klauer:
Superseded by #796

Issue #789, opened 2022-05-13
Error [WinError 10054]


vstadnytskyi-FDA:
I have a caproto server running on Windows 10. On the same machine I am trying to access data via threading client - see below.

I get an error -> [WinError 10054]

Any clue why?
```
In [1]: import logging
   ...: from logging import debug, info, warning, error
   ...:
   ...: logging.getLogger(__name__).addHandler(logging.NullHandler())
   ...:
   ...: from caproto.threading.client import Context
   ...:
   ...: ctx = Context()
   ...: prefix = "camera_FDA"
   ...: image, frameID = ctx.get_pvs(f"{prefix}:image", f"{prefix}:frameID")
Removing <caproto.threading.client.SharedBroadcaster object at 0x00000205892809D0> due to [WinError 10054] An existing connection was forcibly closed by the remote host (10054)
```

klauer:
Can you do a `caproto-get -vvv` with the image/frameID and let us see how it's being discovered?

vstadnytskyi-FDA:
Here is what I am getting on the server side
```

DEBUG:caproto.bcast:VersionRequest(priority=0, version=13)
DEBUG:caproto.bcast:SearchRequest(name='camera_FDA:image', cid=41928, version=13, reply=5)
DEBUG:caproto.bcast:SearchRequest(name='camera_FDA:frameID', cid=41929, version=13, reply=5)
DEBUG:caproto.bcast:VersionResponse(version=13)
DEBUG:caproto.bcast.search:SearchResponse(port=5064, ip='255.255.255.255', cid=41928, version=13)
DEBUG:caproto.bcast.search:SearchResponse(port=5064, ip='255.255.255.255', cid=41929, version=13)
DEBUG:caproto.bcast:VersionRequest(priority=0, version=13)
DEBUG:caproto.bcast:SearchRequest(name='camera_FDA:image', cid=41928, version=13, reply=5)
DEBUG:caproto.bcast:SearchRequest(name='camera_FDA:frameID', cid=41929, version=13, reply=5)
DEBUG:caproto.bcast:VersionResponse(version=13)
DEBUG:caproto.bcast.search:SearchResponse(port=5064, ip='255.255.255.255', cid=41928, version=13)
DEBUG:caproto.bcast.search:SearchResponse(port=5064, ip='255.255.255.255', cid=41929, version=13)
DEBUG:caproto.bcast:VersionRequest(priority=0, version=13)
[...]

PR #794, opened 2022-06-22
FIX/MNT: cleanups and attempts at addressing windows instability

klauer:
Issues
-------
Supersedes #787 
Likely addresses #789 

Notes
-----

- Python 3.6 and Python 3.7 support has been dropped.
- Python 3.8 through Python 3.10 are now targeted for support.

Fixed
-----

- Fixes for all Windows clients, servers, and the repeater relating to PV
  searches. ``recvfrom`` was occasionally raising ``ConnectionResetError``,
  which on Windows only means that: "... a previous send operation resulted in
  an ICMP Port Unreachable message."  This is not fatal in caproto's case and
  can safely be ignored (or worked around, in the case of [asyncio](https://github.com/python/cpython/issues/88906)).
  https://docs.microsoft.com/en-us/windows/win32/api/winsock/nf-winsock-recvfrom
- Fixed pickle/copy for ChannelData subclasses
- Fixed synchronous repeater bind address to INADDR_ANY ("0.0.0.0").  Binding
  to a specific interface meant that on Windows the repeater could not use
  its client-is-alive detection mechanism by binding to the client port as
  the bind would always succeed.

Development
-----------

- Adjusted continuous integration jobs to reflect new supported versions.
- Improved test suite reliability on continuous integration.
- Removed buffer-slicing tools and tests in a refactor to avoid ``sendmsg`` on
  all platforms and async libraries.
- Removed some old windows compatibility patches in
  ``caproto._windows_compat``, and avoided monkey-patching for curio by moving
  its Windows-facing ``curio.run`` modification to
  ``caproto.curio.utils.curio_run``.
- Test suite has some built-in timeout-handling functionality on a
  per-async-library basis, avoiding exiting the entire test suite on Windows
  (due to limitations of ``pytest-timeout``).
- Trio socket handling has been updated to follow better practices for recent
  trio versions.

### DIFF
--- a/caproto/examples/thread_client_simple.py
+++ b/caproto/examples/thread_client_simple.py
@@ -1,6 +1,6 @@
 import time
 
-from caproto.threading.client import SharedBroadcaster, Context
+from caproto.threading.client import Context, SharedBroadcaster
 
 
 def main(pvname1='int', pvname2='str'):
@@ -15,7 +15,7 @@ def main(pvname1='int', pvname2='str'):
     # Some user function to call when subscriptions receive data.
     called = []
 
-    def user_callback(command):
+    def user_callback(sub, command):
         print("Subscription has received data: {}".format(command))
         called.append(command)
 
--- a/caproto/server/common.py
+++ b/caproto/server/common.py
@@ -741,12 +741,18 @@ class Context:
                 bytes_received, address = await udp_sock.recvfrom(
                     MAX_UDP_RECV
                 )
+            except ConnectionResetError:
+                # Win32: "On a UDP-datagram socket this error indicates
+                # a previous send operation resulted in an ICMP Port
+                # Unreachable message."
+                #
+                # https://docs.microsoft.com/en-us/windows/win32/api/winsock/nf-winsock-recvfrom
+                self.log.debug('UDP server recvfrom error')
             except OSError:
                 self.log.exception('UDP server recvfrom error')
-                await self.async_layer.library.sleep(0.1)
-                continue
-            if bytes_received:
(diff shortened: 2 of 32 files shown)


setup.py
```diff
@@ -29,7 +29,9 @@
     'Intended Audience :: Science/Research',
     'Programming Language :: Python',
     'Programming Language :: Python :: 3',
-    'Programming Language :: Python :: 3.6',
+    'Programming Language :: Python :: 3.8',
```
tacaswell, line 32:
```suggestion
    'Programming Language :: Python :: 3.7',
    'Programming Language :: Python :: 3.8',
```

caproto/server/common.py
```diff
@@ -1095,6 +1095,13 @@ async def tcp_handler(self, client, addr):
             while True:
                 try:
                     await circuit.recv()
+                except Exception:
```
tacaswell, line 1111: I think this will full mask the exception below?  I think these need to be re-ordered.

Reading commit-by-commit so you might have already fixed this in the future.

tacaswell: yes you did.

caproto/examples/thread_client_simple.py
```diff
@@ -15,7 +15,7 @@ def main(pvname1='int', pvname2='str'):
     # Some user function to call when subscriptions receive data.
     called = []
 
-    def user_callback(command):
```
tacaswell, line 18: was this just broken before?

klauer: Deprecated API as mentioned in slack

tacaswell (approved):

PR #795, opened 2022-06-22
TST: if the test runs, ignore event loop stoppage

klauer:
(Feel free to point out where I went wrong here; in the meantime this hack may patch it over)


klauer:
Admonishments can come later; tests may pass better than before

PR #796, opened 2022-06-24
REF/ENH: server type annotations + generic support

klauer:
## Status quo: `caproto/master`

The server code is ugly and awful spaghetti

## With this PR

The server code is ugly and awful spaghetti - with type annotations that you can "go to source" on in your favorite IDE

## Compatibility

* Should be almost entirely backward-compatible
* Still Python 3.8+ as in the previous PR
* Fixed bug requiring getter/putter/etc definition to have the same name as the pvproperty
 
## How

* `pvproperty` and record type handling has been changed to support using the `PvpropertyData` and `*Fields` classes directly such that static analyzers can figure out what we're trying to do
* This can be inferred directly from:
    * `pvproperty(dtype=PvpropertyInteger)` (an integer)
    * `pvproperty(dtype=PvpropertyInteger[AiFields])` (an integer with analog input record fields)
    * `pvproperty(dtype=PvpropertyInteger[AiFields], record=AiFields)` (an integer with analog input record fields - required kwarg for the annotations to be useful in the class body)
        *  If that doesn't make sense, try the varieties out in your IDE and see whether or not `prop.fields` works in the class body and whether or not `data.field_inst` works on the instantiated IOC

## Future

- [x] Need to fix the documentation helpers
- [x] Record classes need updating, will be ~in a follow-up PR once I update the conversion scripts~ in this PR; it wasn't a big change

If I ever get around to it (no plans so far):
    * A less magical - but still flexible - `PVGroup` may eventually be a thing.
    * Better ways to separate data definition + implementation


caproto/asyncio/server.py
```diff
@@ -26,9 +26,13 @@ async def wait(self, timeout=None):
         return self.is_set()
 
 
+# Back-compat
+Event = AsyncioEvent
+
+
 class AsyncioAsyncLayer(AsyncLibraryLayer):
     name = 'asyncio'
-    Event = asyncio.Event
```
tacaswell, line 31: This looks like we did not use the custom Event here though?

caproto/server/server.py
```diff
@@ -508,18 +557,49 @@ class PVSpec(namedtuple('PVSpec',
...
     __slots__ = ()
-    default_dtype = int
-
-    def __new__(cls, get=None, put=None, startup=None, shutdown=None,
-                attr=None, name=None, dtype=None, value=None, max_length=None,
-                alarm_group=None, read_only=None, doc=None, fields=None,
-                scan=None, record=None, cls_kwargs=None):
+    default_dtype: ClassVar[type] = int
+
+    get: Optional[Getter]
```
tacaswell, line 562: I think p310 is when we get 

```
    get: Getter | None
```

caproto/asyncio/server.py
klauer: Right, that was a mistake at the time of the asyncio server PR.

The sole difference here is just that the `timeout` kwarg will now actually async-lib agnostic (between curio, trio, and asyncio).

caproto/server/server.py
klauer: That'll be nice!

klauer:
Ping @tacaswell

danielballan:
Woah, I was really on vacation while this landed and missed it. Nice!

doc/source/install.rst
```diff
@@ -126,43 +126,3 @@ To build the documentation:
 .. code-block:: bash
 
     make -C doc html
-
-Installation on iOS
-===================
-
-`Pythonista <http://omz-software.com/pythonista/>`_ brings support for Python
```
tacaswell, line 133: Pity that the (commercial) upstream seems to have not updated from 2020 and no longer provides a version of Python we support).

tacaswell:
YOLO

PR #812, opened 2023-04-27
TernaryArrayIOC

gwbischof:
A TernaryDevice is a device with 3 states.  Many components at a beamline have this behavior.  For example, a valve, has an open and close command, a closed state, an open state, and an in-between state.

The TernaryArrayIOC is an IOC that has an array of TernaryDevices.
An example of a device that has an array of ternary devices is an Attenuator, which is an array of filters.

I think an interesting thing about this example is the dynamic pv creation.
Also, I think this is useful to test Ophyd devices like the Attenuator at CMS.

### DIFF
--- a/caproto/ioc_examples/ternaryarray.py
+++ b/caproto/ioc_examples/ternaryarray.py
@@ -52,7 +52,6 @@ class TernaryArrayIOC(PVGroup):
 
     def __init__(self, count=10, *args, **kwargs):
         self._devices = [TernaryDevice() for i in range(count)]
-        super().__init__(*args, **kwargs)
 
         # Dynamically setup the pvs.
         for i in range(count):
@@ -70,7 +69,11 @@ class TernaryArrayIOC(PVGroup):
             # Create the readback scan.
             partial_scan = partial(self.general_scan, i)
             partial_scan.__name__ = f"scan{i}"
-            getattr(self, f'device{i}_rbv').scan(partial_scan, period=0.1)
+            getattr(self, f'device{i}_rbv').scan(period=0.1)(partial_scan)
+
+        super().__init__(*args, **kwargs)
+        # Register the late pvs.
+        self.pvdb.update(PVGroup.find_pvproperties(self.__dict__))
 
     async def general_putter(self, index, group, instance, value):
         if value:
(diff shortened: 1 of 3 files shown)


caproto/ioc_examples/ternaryarray.py
```diff
@@ -0,0 +1,97 @@
...
+
+    async def general_putter(self, index, group, instance, value):
+        if value:
+            self._devices[index].set()
+        else:
+            self._devices[index].reset()
+
+    async def general_scan(self, index, group, instance, async_lib):
+        # A hacky way to write to the pv.
+        await self.pvdb[f'{self.prefix}device{index}_rbv'].write(self._devices[index].state)
```
gwbischof, line 86: I think that there might be a way to get rid of this hack. Any ideas?

caproto/ioc_examples/ternaryarray.py
```diff
@@ -0,0 +1,143 @@
...
+    ----------
+    count: integer
+        The number of devices in the array.
+    """
+
+    def __init__(self, count=10, *args, **kwargs):
+        self._devices = [TernaryDeviceSim() for i in range(count)]
+
+        # Dynamically setup the pvs.
+        for i in range(count):
```
tacaswell, line 55: Do this as a helper function that uses `type` to define a new class.  Going up (down?) an abstraction layer will simplify this a a bunch.

caproto/ioc_examples/ternaryarray.py
```diff
@@ -0,0 +1,143 @@
...
+            # Create the setpoint pv.
+            setattr(
+                self, f"device{i}", pvproperty(value=0, dtype=int, name=f"device{i}")
```
tacaswell, line 58: How does this translate {True, False, None} to an integer?

gwbischof: I'm going back and forth a bit it figure out the best way to do this.

I was thinking that I would be able to make TernaryDevice general, so that it can work with any device that has the same behavior.  So that it doesn't make any assumptions about pv names or the strings that get returned by status pv.
Then I was planning on subclassing TernaryDevice, to make a class called Filter which includes some of specifics about the cms filters, like the strings that get returned by the status PV.
Then I was planning on making a device that has an array of Filters for the cms attenuator.

Anyway, I'll have to work out this detail before it is ready to merge.

tacaswell: Can we use enums for this?

PR #815, opened 2023-05-17
CI: move azure pipelines to Ubuntu 22.04

klauer:
18.04 was removed in December 2022 per https://learn.microsoft.com/en-us/azure/devops/release-notes/2022/sprint-208-update

If this upgrade isn't easy, I'm going to just nuke Azure Pipelines entirely from this repository because it's failing our test suite unnecessarily


klauer:
Yep, let's merge this in. Easier than expected (fortunately!)

PR #814, opened 2023-05-17
FIX: server subscriptions affecting each other

klauer:
Closes #813 

See the linked issue for further details.

Tested and found to work locally:
* The array slice filter (with two separate monitors)
* The deadband filter

This is fundamental code to caproto's server mechanism, so a detailed/thorough review would be wise and appreciated.

PR #816, opened 2023-05-18
FIX: issue 797 - subscription queue loss with eventsoff/eventson

### LINKED ISSUE #797: High load. Dropped 4 responses.
I am develping software IOC using captoro.
A client software is built wit CSS.
When the OPI is started or refreshed, the software IOC show the following message:

[W 11:45:20.034       common:  385] High load. Dropped 4 responses.

It looks some widgets on the OPI failed to update values.
I hope not to drop reseponses, even if responses are slow.

klauer:
## Context

(Builds on #814 as it refactored subscription code; that must be merged first)

Closes issue #797 

Reproduced with: https://github.com/klauer/issue_797_caproto_subs
Server: `support-ioc.py` (run this first)
Client: `client_issue_797.py` (run this second)

The final result should indicate that each PV saw one subscription total.
caproto master will range from 1 to 19 subscriptions on average.
This PR should bump that up to the expected 20.

## The fix

This is at least a proof-of-concept method to fix #797. I'm not really sure if it's the best way (if there is one).

Subscription values - generated once and held by way of weakref - were getting cleaned up before the queue had a chance to send them.

I think the current implementation attempts to be smart and save effort in the end, but the complexity in doing so introduces potential for confusing errors like this one.

It may indicate that removing `weakref` entirely could be preferable (at risk of ballooning memory due to greedy clients) - but subscriptions are supposed to be lossless anyway, right?

Regardless, I took the path of making a smaller change rather than removing weakref entirely. This PR says "queue up the latest updates when events are off, but if they get garbage collected before events turn back on again, make note of that and redo the subscription - by reading from the database entry directly - when events turn on"

Thoughts? Anyone? ... Bueller?

### DIFF
--- a/caproto/_utils.py
+++ b/caproto/_utils.py
@@ -1,9 +1,12 @@
 # This module includes all exceptions raised by caproto, sentinel objects used
 # throughout the package (see detailed comment below), various network-related
 # helper functions, and other miscellaneous utilities.
+from __future__ import annotations
+
 import argparse
 import array
 import collections
+import copy
 import enum
 import functools
 import inspect
@@ -20,9 +23,15 @@ import typing
 import weakref
 from collections import namedtuple
 from contextlib import contextmanager
+from typing import Iterable
 from warnings import warn
 
+from ._dbr import SubscriptionType
 from ._version import get_versions
+
+if typing.TYPE_CHECKING:
+    from .server.common import Subscription
+
 
 __version__ = get_versions()['version']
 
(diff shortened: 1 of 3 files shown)


danielballan:
Since we wrote that I have cooled on using weakref for this kind of thing. It is a little too implicit and nondeterministic. I was also influenced by https://crawshaw.io/blog/tragedy-of-finalizers

I think this minimal-diff approach is good for now, if patience and free time are dwindling. There is something circuitous about this (on removal, add it back…) that will be a good opportunity for a future refactor.

klauer:
`weakref` still has somewhat of a place in the subscription mechanism - at least it makes more sense its usage in #817.

I'm leaning toward the alternate PR and have submitted a test suite case for the issue there. If tests pass on CI, I intend to merge that one.

klauer:
Closing in favor of #817

PR #819, opened 2023-05-19
ENH: subscription backlog configuration and better-tuned settings

### LINKED ISSUE #780: Document meaning of "High load. Batched 2 commands" and warn only if above a threshold
I am getting this message but I do not how to make sense out of it. See below. Is it something to do with sending too much data or sending to many updates to PVs? 

```
[I 11:50:39.604       common:  390] High load. Batched 2 commands (168B) with 0.0003s latency.
[I 11:50:49.370       common:  390] High load. Batched 2 commands (168B) with 0.0033s latency.
[I 11:50:52.602       common:  390] High load. Batched 4 commands (416B) with 0.0005s latency.
[I 11:50:58.699       common:  390] High load. Batched 2 commands (168B) with 0.0002s latency.
[I 11:51:00.885       common:  390] High load. Batched 2 commands (168B) with 0.0008s latency.
[I 11:51:01.226       common:  390] High load. Batched 4 commands (416B) with 0.0002s latency.
[I 11:51:09.341       common:  390] High load. Batched 2 commands (168B) with 0.0006s latency.
[I 11:51:09.428       common:  390] High load. Batched 2 commands (208B) with 0.0012s latency.
[I 11:51:11.877       common:  390] High load. Batched 2 commands (168B) with 0.0124s latency.
[I 11:51:12.216       common:  390] High load. Batched 4 commands (416B) with 0.0003s latency.
[I 11:51:20.059       common:  390] High load. Batched 2 commands (168B) with 0.0008s latency.
[I 11:51:23.274       common:  390] High load. Batched 2 commands (168B) with 0.0002s latency.
[I 11:51:23.656       common:  390] High load. Batched 4 commands (416B) with 0.0005s latency.
[I 11:51:28.470       common:  390] High load. Batched 2 commands (168B) with 0.0006s latency.
[I 11:51:28.681       common:  390] High load. Batched 2 commands (168B) with 0.0009s latency.
[I 11:51:30.737       common:  390] High load. Batched 2 commands (168B) with 0.0143s latency.
[I 11:51:33.044       common:  390] High load. Batched 2 commands (168B) with 0.0069s latency.
[I 11:51:33.393       common:  390] High load. Batched 4 commands (416B) with 0.0003s latency.
[I 11:52:34.463       common:  390] High load. Batched 4 commands (416B) with 0.0002s latency.
[I 11:53:36.582       common:  390] High load. Batched 4 commands (416B) with 0.0002s latency.
[I 11:54:34.552       common:  390] High load. Batched 4 commands (416B) with 0.0002s latency.
```

### LINKED ISSUE #791: Subscription backlog memory usage too high for PVs storing large arrays
Hi All,

I have created a caproto server that has to PVs:
- image aray
- update rate

and a GUI(pyepics based) that show the image array as a BitMap

when I start my GUI, I see slow increase in RAM allocation. So my first thought was that there is a problem with GUI, but if I look at which process takes more RAM I see that the caproto server starts eating more and more RAM

image

If I turn off GUI or server, the RAM allocation goes back to baseline. This is s repeatable behavior. It almost seems like that images are getting queued up and they are not getting cleared.

see below for codes.

Server code
https://github.com/vstadnytskyi/caproto-sandbox/blob/master/caproto_sandbox/external_screen/bitmap_image_server.py

```
#!/usr/bin/env python3
"""
caproto server that generates a 8-bit bitmap image of given size
"""

from logging import debug,info,warning,error
#Input/Output server library
from caproto.server import pvproperty, PVGroup, SubGroup, ioc_arg_parser, run
#from ubcs_auxiliary.threading import new_thread
from numpy import nan, zeros, int16, random
width = 1920 #1920 #int(1920/2)
height = 1080 #1080#int(1080/2)

class Server(PVGroup):
    
    image = pvproperty(value=zeros((height,width,3)).flatten(), dtype = int, max_length = width*height*3*10)
    dt = pvproperty(value = 1)

    @image.startup #this fubction will be executed on instantiaton
    async def image(self, instance, async_lib):
        pos_w = 0
        pos_h = 0
        self.new_arr = zeros((height,width,3))
        while True:
            #new_arr = random.randint(256, size=(3, 1000, 500)).flatten()

            self.new_arr = self.new_arr*0
            #new_arr[int(height/2),:,1] = 255
            self.new_arr[:,pos_w,:] = 255
            self.new_arr[pos_h,:,:] = 255
            if pos_w == (width-1):
                pos_w = 0
            else:
                pos_w += 1
            if pos_h == (height-1):
                pos_h = 0
            else:
                pos_h += 1
            await self.image.write(value=self.new_arr.flatten())
            # Let the async library wait for the next iteration
            await async_lib.library.sleep(self.dt.value)
    #@image.putter
    #async def image(self, instance, value):

if __name__ == '__main__':
    # The Record Name is specified by prefix
    prefix = 'BITMAP_IMAGE:'
    from pdb import pm
    from tempfile import gettempdir
    import logging
    print(gettempdir()+'/{}.log'.format(prefix))
    logging.basicConfig(filename=gettempdir()+'/{}.log'.format(prefix),
                        level=logging.INFO, format="%(asctime)s %(levelname)s: %(message)s")

        default_prefix=prefix,
        desc='description')

    io_server  = Server(**ioc_options)

    #start async caproto server IO
    run(io_server.pvdb, **run_options)
```

GUI Code
https://github.com/vstadnytskyi/caproto-sandbox/blob/master/caproto_sandbox/external_screen/bitmap_image_gui_full_screen.py

```
#!/usr/bin/python
# -*- coding: utf-8 -*-

import wx
import epics
import epics.wx
from logging import debug,warn,info,error
from pdb import pm

import os
os.environ["EPICS_CA_MAX_ARRAY_BYTES"] = '48000096' #there is a limit of how big PV can be on the client size. This overrrides that limitation.

from icarus_nmr.pyepics import PVImage

__version__ = "0.0.0" #initial

class MyPanel(wx.Panel):
    """"""

    def __init__(self, parent):
        """Constructor"""
        wx.Panel.__init__(self, parent)

        self.Bind(wx.EVT_KEY_DOWN, self.onKey)
        self.SetBackgroundColour((0, 0, 0))

    def onKey(self, event):
        """
        Check for ESC key press and exit is ESC is pressed
        """
        key_code = event.GetKeyCode()
        if key_code == wx.WXK_ESCAPE:
            self.GetParent().Close()
        else:
            event.Skip()
class MyFrame(wx.Frame):
    """"""

    def __init__(self):
        """Constructor"""   
        wx.Frame.__init__(self, None, title="Test FullScreen")
        self.panel = MyPanel(self)
        self.initialize_GUI()
        self.SetBackgroundColour(wx.Colour(255,255,255))
        self.ShowFullScreen(True)

    def initialize_GUI(self):
        """
        """

        width = 1920#1920 #int(1920/2)
        height = 1080 #1080#int(1080/2)

        sizer = wx.GridBagSizer(hgap = 10, vgap = 10)
        self.labels ={}
        self.fields = {}
        self.sizers = {}
        main_sizer = wx.BoxSizer(wx.VERTICAL)
        topSizer = wx.BoxSizer(wx.HORIZONTAL)

        self.sizers['image'] = wx.BoxSizer(wx.VERTICAL)
        
        self.fields['image']  = PVImage(self.panel, pv='BITMAP_IMAGE:image', im_size = (width,height))
        self.sizers['image'] .Add(self.fields['image']  , 0)

        self.panel.SetSizer(topSizer)

if __name__ == '__main__':
    from pdb import pm
    import logging
    from tempfile import gettempdir

    app = wx.App(False)
    frame = MyFrame()
    frame.SetPosition((-800,-1000))
    frame.Maximize()
    app.MainLoop()
```

klauer:
# Context

Closes #780 
Closes #791 

I think this is the last PR before v1.1.0, unless I see other low-hanging fruit

## Changes

* Adds `max_subscription_backlog` on a per-`ChannelData` instance.
* This is configurable as a global constant/environment variable: `CAPROTO_MAX_SUBSCRIPTION_BACKLOG`
   * There are an additional set of thresholds that allow _configurable_ automatic reduction of the backlog on a per PV basis, if they are too large. This avoids the situation where a single waveform being monitored could cause an OOM situation
* Made almost all "tweakable" constants environment variables. Added a minimal set of documentation for each - hey, at least they're configurable


doc/source/environment_variables.rst
```diff
@@ -69,7 +102,37 @@ Environment Variables
...
+     - 0.03
+     - Warn the user if packets are delayed by more than this amount: 30ms.
+       Set to 0 to disable the warning entirely.
+   * - CAPROTO_SERVER_SUB_BATCH_THRESH
+     - 2 ** 16
+     - When a batch of subscription updates has this many bytes or more, send
+       it.
+   * - CAPROTO_SERVER_MAX_LATENCY
+     - 1
+     - Tune this to change the max time between packets. If it's too high, the
```
tacaswell, line 128: units?  I assume seconds?

tacaswell:
This all seems reasonable, however it looks like this broke a bunch of time sensitive tests?

klauer:
> This all seems reasonable, 

👍 

> however it looks like this broke a bunch of time sensitive tests?

Ugh, I wouldn't have expected it to break anything. I'll dig in and see what happened...

doc/source/environment_variables.rst
klauer: Good point - I included units in all relevant ones in the docs and switched up env var names to reflect that

klauer:
Looks like one random test failure. I think the issue boiled down to a mistaken `int(env_var)` instead of `float(env_var)`, effectively breaking some aspect of CA by zeroing an important parameter out.

I expect people won't be tweaking these settings unless they run into issues. It's then up to the user to figure out what a good value is, with the knowledge that caproto will dutifully do what you ask and may fall over in the process.

PR #820, opened 2023-05-19
CI: fix GitHub actions pypi uploading workflow

### LINKED ISSUE #762: Fix PyPI uploading
For next release, the following should be something like:

https://github.com/caproto/caproto/blob/1e37e13883bcd97a679339aaa3b31fd081dba0aa/.github/workflows/pypi.yml#L35

`twine upload dist/*`

klauer:
Closes #762


klauer:
GHA jobs passing with the install change. We'll see what happens when the tag comes around.

PR #825, opened 2023-10-23
Reuse existing context if none is specified

### LINKED ISSUE #824: Resource leak in caproto.threading.pyepics_compat
At the moment, `caproto.threading.pyepics_compat.caget_many` creates a new CA context every time the function is called, unless a context object is passed explicitly. This is due to the following code in `caget_many`:

```py
    if context is None:
        context = Context(PV._default_context.broadcaster)
```

This means that when this function is called repeatedly, many context objects are created, which potentially leads to many connections being established to the same CA server, which can result in resource exhaustion on the server (this is how we noticed this behavior).

I believe that the fix is rather simple: The default context object, that is also used by the other functions in this module should also be used by the`caget_many` function. For example, `get_pv` uses the following code:

```
    if context is None:
        context = PV._default_context
```

I can open a PR for this fix, I just wanted to verify first that this is indeed a bug not intended behavior for some reason.

smarsching:
`caproto.threading.pyepics_compat.caget_many` inadvertently created a new CA context every time this function was called without an explicitly specified context. With this PR, it instead reuses the default context, like the other functions in the module do.

This closes #824.

### DIFF
--- a/caproto/threading/pyepics_compat.py
+++ b/caproto/threading/pyepics_compat.py
@@ -1032,7 +1032,7 @@ def caget_many(pvlist, as_string=False, count=None, as_numpy=True, timeout=5.0,
     as possible to fetch many values.
     """
     if context is None:
-        context = Context(PV._default_context.broadcaster)
+        context = PV._default_context
 
     pvs = context.get_pvs(*pvlist)
 


klauer (approved):
Thanks for reporting and fixing this!

(CI failures are unrelated to your changes - caproto is a _bit_ behind in maintenance)

PR #837, opened 2024-02-12
WIP: split the argparser logic to one more level

tacaswell:
Currently we have two levels to the factory function to generate the CLI interface for caproto IOCs

- ioc_arg_parser -> all-in-one that gives you two dictionaries
- template_arg_parser -> gives you a parser and a function to extract those dictionaries

The first is useful for reducing boiler plate in servers, the second is useful if you need to inject some additional parameters to be extracted from the CLI.

This adds another layer which giver a ArgParser will add all of the standard flags which is useful when you have an existing CLI tool that you want to add caporto features to.


wigging:
Can you provide a basic example of how these new features are implemented by a user? How would `template_arg_parser` be used for the example shown below?

```python
from caproto.server import PVGroup, ioc_arg_parser, pvproperty, run

class SimpleIOC(PVGroup):
    A = pvproperty(value=1, doc="An integer")
    B = pvproperty(value=2.0, doc="A float")
    C = pvproperty(value=[1, 2, 3], doc="An array of integers (max length 3)")

if __name__ == "__main__":
    ioc_options, run_options = ioc_arg_parser(default_prefix="simple:", desc="Example")
    ioc = SimpleIOC(**ioc_options)
    run(ioc.pvdb, **run_options)
```

wigging:
Adding a log level parameter to the `run` function would be useful for situations when not using the `ioc_arg_parser` approach.

```python
from caproto.server import PVGroup, pvproperty, run

class SimpleIOC(PVGroup):
    A = pvproperty(value=1, doc="An integer")
    B = pvproperty(value=2.0, doc="A float")
    C = pvproperty(value=[1, 2, 3], doc="An array of integers (max length 3)")

if __name__ == "__main__":
    ioc = SimpleIOC(prefix="simple")
    run(ioc.pvdb, loglevel="info")
```

tacaswell:
https://github.com/caproto/caproto/blob/46020e4816c7dbe8ccae9b7f0af362ce9a7f4f07/caproto/ioc_examples/chirp.py#L52-L74 is one of a couple of existing examples of using `template_arg_parser`.

To use the new one it would be something like

```python

...  # other code
parser = ...  # where ever you are getting it from now
split_function = extend_arg_parser(parser)  # mutates in place!
... # other code
args = parser.parse_args()
... # other code
ioc_options, run_options = split_args(args)
ioc = SimpleIOC(**ioc_options)
run(ioc.pvdb, **run_options)

```

caproto/server/server.py
```diff
@@ -2102,21 +2103,26 @@ class _ReadWriteSubGroup(PVGroup, Generic[T_Data, T_RecordFields]):
...
     default_prefix: str,
     argv: Optional[List[str]] = None,
     macros: Optional[Dict[str, str]] = None,
-    supported_async_libs: Optional[List[str]] = None
-) -> Tuple[argparse.ArgumentParser, Callable]:
+    supported_async_libs: Optional[List[str]] = None,
+) -> Callable:
     """
-    Construct a template arg parser for starting up an IOC
+    Extend an exsiting arg parser for standard flags.
```
klauer, line 2115:
```suggestion
    Extend an existing arg parser with standard caproto IOC flags.
```

caproto/server/server.py
```diff
@@ -2210,7 +2211,56 @@ def split_args(args) -> Tuple[Dict[str, Any], Dict[str, Any]]:
...
+    supported_async_libs: Optional[List[str]] = None,
+) -> Tuple[argparse.ArgumentParser, Callable]:
+    """
+    Construct a template arg parser for starting up an IOC
+
+    Parameters
+    ----------
+    description : string
+        Human-friendly description of what that IOC does
+    default_prefix : string
```
klauer, line 2232:
```suggestion
    description : str
        Human-friendly description of what the IOC does
    default_prefix : string
```

caproto/server/server.py
```diff
@@ -2210,7 +2211,56 @@ def split_args(args) -> Tuple[Dict[str, Any], Dict[str, Any]]:
...
+    args : list, optional
+        Defaults to sys.argv
+    macros : dict, optional
+        Maps macro names to default value (string) or None (indicating that
+        this macro parameter is required).
+    supported_async_libs : list, optional
+        "White list" of supported server implementations. The first one will
```
klauer, line 2239:
```suggestion
        Allowed list of supported server implementations. The first one will
```

klauer (commented):
Seems pretty reasonable; what's holding this up?

caproto/server/server.py
tacaswell, line 2232:
```suggestion
    default_prefix : str
```

tacaswell:
@toqduj @codedump thoughts on this extension as people who have written IOCs?

codedump:
Oooff... it's complicated :-) 

I'm not sure my answer will be any help because I've actually never used the CAproto application / CLI-building layers. But here we go!

**Short answer:** I think that having the possibility to inject IOC-relevant options into an existing arg-parser is a very good thing. So, +1.

**Long answer:**

We're actually don't bother implementing command-line arguments anymore. We did in the past. I've even attempted to write our own framework to deal with all the boilerplate "back in the day", when we still used mostly `pythonSoftIOC`. It was *not* a complete disaster. It kind-of worked. But we've all but abandoned it because it taught us two valuable lessons:
  - balancing the design the right way is *hard* -- one either ends up baking lots of assumptions about the application (which turn out to be true only in half the cases), or ends up with a CLI structure that's too generic to actually be user-friendly
  - on the IOC/application programmer side, one tends to still end up with lots of boilerplate code -- it's just that the nature of the boilerplate is shifting, from plain old arg-parse, to the new-and-not-so-much-improved IOC arg-framework

Together with the next point, it eventually felt like a battle not worth fighting: we're making a point of deploying all of our IOCs as Docker containers. And what is generally true for deployment of Docker applications in other frameworks is *even more* true for our particular setup: from a system administrator's point of view, it's a lot easier to customize the deployment and behavior of applications using environment variables than using command-line arguments.

For those interested in details: we're using Podman + [Quadlets](https://docs.podman.io/en/latest/markdown/podman-systemd.unit.5.html), which means we're launching the IOCs using Systemd-Unit files. We now can supply pretty generic `<ioc-name>.container` files for every IOC we build, and customization is done by using [systemd drop-in files](https://www.man7.org/linux/man-pages/man5/systemd.unit.5.html). This makes it very easy to override or add an `Enviroment=...` section to change parameters. But also makes it very hard to override the `Exec=...` section without breaking stuff, which we'd have to do if customization were only possible through CLI arguments.

Code-wise, using env-vars allows us to skip lots of the boilerplate code, too. So I'm guessing it's a win-win... the only cost being that the IOC application, when used manually from the command line, has an ugly syntax, I guess.

That said, we now come full-circle to the "short answer": some options aren't about IOC customization, but about the application structure in itself. For instance, we've had an application (proof-of-concept) which could either display a result using Matplotlib, or publish it as a PV: `application.py ... --display` vs `application.py ... --publish-as MY:SHINY:IOC`. 

This part would certainly be easier to implement if there was a possibility to inject IOC-relevant options into the arg-parser, and while doing this env-var based would be possible, it doesn't bring any advantage over a CLI-style version. Then again, maybe this style of application would benefit more from a specific base class, which truly cuts down on boilerplate by *not* being generic -- making the class impossible to use in other situations, but truly easy to use for this style of applications (?).

I also fully agree with @tacaswell 's take of not mixing logging/loglevels into this too hard. One thing we're aiming at is supporting [OpenTelemetry](https://opentelemetry.io/)-based logging and tracing, so we'd have to implement much of the log-preparation boilerplate anyway. We wouldn't benefit from CAproto doing it for us.

toqduj:
I don't fully understand what the problem is that is being solved here. When we have IOCs with special CLI parameters, we build it up like this: 
```python 
if __name__ == '__main__':
    parser, split_args = template_arg_parser(
        default_prefix='mc0:', # motor controller at address zero. 
        desc=dedent(TrinamicIOC.__doc__),
        supported_async_libs=('asyncio',)
        )
    parser.add_argument(
        '--configfile', 
        help='Path to the board- and axes YAML configuration file', 
        required=True, 
        type=validate_file
        )
    args = parser.parse_args()
    ioc_options, run_options = split_args(args)

    ioc = TrinamicIOC(config_file=args.configfile, **ioc_options)
    run(ioc.pvdb, **run_options)
```

tacaswell:
The case is where you have some existing code you want to extend with caproto and already have an instantiated parser leaving you with the option of re-implementing either caprotos or the other programs flags.  In this case we are trying to be the good citizen and provide a function to add our flags to the existing parser.

This could very well be a super niche need.   I no longer remember exactly what inspired me to write this.

PR #846, opened 2024-04-16
FIX: convert arrays to native endianess in pyepics_compat

tacaswell:
Alternative to https://github.com/bluesky/ophyd/pull/1186

### DIFF
--- a/caproto/tests/test_pyepics_compat2.py
+++ b/caproto/tests/test_pyepics_compat2.py
@@ -57,6 +57,7 @@ def test_put_empty_list(context, ioc):
 
     pv.put([], wait=True)
     ret = pv.get(use_monitor=False)
+    assert ret.dtype.isnative
     assert list(ret) == []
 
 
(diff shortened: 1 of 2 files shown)


cjtitus:
I tested this and it works. Along the way, it illuminated for me why this only popped up in containers. It turns out that Ophyd is falling back on Caproto as the control layer, presumably because I haven't installed pyepics. And _caproto_shim.py does indeed call the pyepics_compat in Ophyd.

So presumably the _caproto_shim in Ophyd never gets used at NSLS-II because we always run in these massive Conda environments that have pyepics, but this could actually pop up somewhere else if someone's installing a more minimal environment from scratch.

klauer (approved):
Seems good to me

Where'd you find a big endian device?

tacaswell:
conflict was trivial (two tests added at the bottom of the file).

tacaswell:
Merging on @klauer 's approval.

PR #852, opened 2024-11-15
Enh/no search mirror

tacaswell:
Still in develpoment, but opening early for feedback.


tacaswell:
I need to pull a few bug-fixes from using this in anger back.

PR #862, opened 2025-04-01
Make PV._default_context a public class method

thopkins32:
Currently, doing
```
PV._default_context
```
fails because it is actually an instance method rather than a class method. It is used this way in a couple of places:
- https://github.com/caproto/caproto/blob/284568779305335fed069459999a65e933cff93f/caproto/threading/pyepics_compat.py#L1043
- https://github.com/bluesky/ophyd/blob/main/ophyd/_caproto_shim.py#L202

This PR makes it a proper class method with caching.

Here is a small script I wrote to test this:
```python
import time
import functools


@functools.lru_cache(1)
def context():
    time.sleep(5)
    return Context({"name": "test"})

class Context:
    def __init__(self, context):
        self.context = context

    def __str__(self):
        return f"Context: {self.context}"


class TestClass:

    @classmethod
    @property
    @functools.cache
    def _default_cls_context(cls):
        return context()

    @functools.cached_property
    def _default_inst_context(self):
        return context()

    def __init__(self, name):
        self.name = name

    def __str__(self):
        return f"TestClass: {self.name}"

if __name__ == "__main__":
    print("First----------------------------------")
    print(TestClass._default_cls_context.context)
    print("Second----------------------------------")
    print(TestClass._default_cls_context.context)
    print("Done.")

```

### DIFF
--- a/caproto/threading/pyepics_compat.py
+++ b/caproto/threading/pyepics_compat.py
@@ -195,7 +195,6 @@ class PV:
                'put_complete')
 
     @classmethod
-    @functools.cache
     def default_context(cls):
         return _make_context()
 
@@ -1041,7 +1040,7 @@ def caget_many(pvlist, as_string=False, count=None, as_numpy=True, timeout=5.0,
     as possible to fetch many values.
     """
     if context is None:
-        context = _make_context()
+        context = PV.default_context()
 
     pvs = context.get_pvs(*pvlist)
 
--- a/caproto/threading/pyepics_compat.py
+++ b/caproto/threading/pyepics_compat.py
@@ -196,7 +196,7 @@ class PV:
 
     @classmethod
     @functools.cache
-    def _default_context(cls):
+    def default_context(cls):
         return _make_context()
 
     def __init__(self, pvname, callback=None, form='time',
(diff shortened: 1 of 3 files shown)


thomashopkins32:
TIL: Chaining `@classmethod` and `@property` is deprecated since Python 3.11 and removed in Python 3.13.

I'll try to make a better change for this.

caproto/threading/pyepics_compat.py
```diff
@@ -194,8 +194,9 @@ class PV:
                'upper_warning_limit', 'upper_ctrl_limit', 'lower_ctrl_limit',
                'put_complete')
 
-    @functools.cached_property
-    def _default_context(self):
+    @classmethod
+    @functools.cache
```
tacaswell, line 198: Do we need the cache here as `_make_context()` is already cached.

thopkins32:
Most recent commit reverts the changes in #864 but I believe they are equivalent now.

tacaswell (approved):

thopkins32:
@tacaswell Just had to update the unit tests since this is a classmethod now. Let me know if it's okay to merge. Thanks!

PR #869, opened 2025-09-18
Safer expand_macros

cjtitus:
Updates expand_macros to only attempt macro substitution if a non-empty dictionary is passed.

In the case where macros are not passed, calling format carries a substantial risk of error, if the string contains instances of "{" or "}" which are not paired, which can easily happen in nested PVGroups that are attempting to follow NSLS-II standard PV name formats.

Needs testing, specifically with actual macros. I have tested the case where you don't want to pass in macros.

We could discuss adding a command-line switch for this, to make it opt-in, and eliminate all risk of modifying the current behavior of Caproto. But I can't imagine what the case is where you're not passing any macros, but `pv.format(**{})` does something that you want.

### DIFF
--- a/caproto/server/server.py
+++ b/caproto/server/server.py
@@ -1808,9 +1808,13 @@ class pvfunction(SubGroup):
 
 
 def expand_macros(pv, macros):
-    'Expand a PV name with Python {format-style} macros'
-    return pv.format(**macros)
-
+    '''If macros are given, expand a PV name with Python {format-style} macros. 
+    If macros is an empty dictionary, return the PV unchanged.
+    '''
+    if macros:
+        return pv.format(**macros)
+    else:
+        return pv
 
 class PVGroupMeta(type):
     'Metaclass that finds all pvproperties'


tacaswell:
I am torn, on one hand, if you are not doing doing the macro expansion, then there is no reason for you to pay the cost of escaping your `{{` so that format works, but on the other hand this does put a step in where you go from not caring to caring and if you are trying to use macros, but don't pass them now your PVs will silently "work" with the `{}` left in them.

On net I'm at 👍🏻 , but a cli flag of "--force-macros" may be good.

cjtitus:
RE: the cost of escaping `{{` -- in my use, I have found that for nested PVGroups, it can be nearly impossible to figure out how to do the escaping properly. Because different levels get expand_macros called a different number of times. so the top-level needs `{{`, the next level needs `{{{{`, etc. I have seen, in the wild, repositories that attempt to automatically determine how many `{{`s need to be added. If you have unbalanced `{`'s, it gets even worse (which some of our actual PVs do!!)

Let me know how you would want a cli flag to be structured -- default to trying to expand macros, with an option to turn off? Or vice-versa?

tacaswell:
Hmm, if the problem is multiple applications, I think the right path is to add a `--no-macros` flag that just turns them off all together as if it is un-escapable then I think it is better to fully skip the feature than to have some implicit sometimes skipping (and maybe raise if the flag is set and a macro is setup?).

cjtitus:
I'd be happy to add a --no-macros flag. It'll be a bit more work to thread through, but shouldn't be terrible.

cjtitus:
Oh, actually -- I have no idea how you'd do that with a command-line flag, as I'm pretty sure expand_macros gets called before the command line flags are even parsed. It's done at class-definition time, right?

I could make it an option to the PV classes, so that you'd build it into your classes, i.e,
```python
class CoolPVs(PVGroup):
    pv1 = pvproperty(name="{}{{{}{}}", expand_macros=False)
```

That's more annoying, but gives you more granular control, and again, anything is better than the "how many `}}`s do I need to add today" game.

PR #872, opened 2026-06-16
Allow longer PV names

smarsching:
While the EPICS IOC code in EPICS Base limits record names to 60 bytes by default, this limit does not apply to the Channel Access protocol (and even in EPICS Base, it can easily be lifted by changing a preprocessor macro).

In fact, the CA client in EPICS Base allows for search requests with much larger channel names (usually up to 999 characters long).

This patch updates the code in caproto so that it uses the same limits that are used by EPICS Base, making it compatible with a wider variety of Channel Access servers and the names that they use.

For those who are interested in the details:

EPICS Base defines `MAX_UDP_SIZE` as 1024 in `caProto.h`. In `nciu.cpp`, it checks that names (including the terminating null byte) do not exceed `MAX_UDP_SIZE` minus the size of the `caHdr` data structure (which is 24), so the effective limit is 1000 bytes.

### DIFF
--- a/caproto/_commands.py
+++ b/caproto/_commands.py
@@ -31,7 +31,7 @@ import _ctypes
 
 from . import _dbr as dbr
 from ._backend import backend
-from ._constants import DO_REPLY, MAX_RECORD_LENGTH, NO_REPLY
+from ._constants import DO_REPLY, NO_REPLY
 from ._dbr import (DBR_INT, DBR_TYPES, MAX_STRING_SIZE, AccessRights,
                    ChannelType, float_t, native_type, short_t, special_types,
                    ushort_t)
@@ -58,7 +58,7 @@ from ._status import eca_value_to_status, ensure_eca_value
 from ._utils import (CLIENT, NEED_DATA, REQUEST, RESPONSE, SERVER,
                      CaprotoNotImplementedError, CaprotoTypeError,
                      CaprotoValueError, RemoteProtocolError, ValidationError,
-                     ensure_bytes)
+                     name_to_bytes, padded_len, padded_string_payload)
 
 __all__ = ('AccessRightsResponse', 'ClearChannelRequest',
            'ClearChannelResponse', 'ClientNameRequest',
@@ -127,11 +127,6 @@ def from_buffer(data_type, data_count, buffer):
     return md_payload, data_payload
 
 
-def padded_len(s):
-    "Length of a (byte)string rounded up to the nearest multiple of 8."
-    return 8 * ((len(s) + 7) // 8)
-
-
 def pad_buffers(*buffers):
     '''Get a bytestring for padding a concatenated set of buffers
 
@@ -146,12 +141,6 @@ def pad_buffers(*buffers):
     unpadded_size = sum(bytelen(buf) for buf in buffers)
     pad_buffer = _pad_buffer[unpadded_size % 8]
     return unpadded_size + len(pad_buffer), pad_buffer
-
-
-def padded_string_payload(payload):
-    byte_payload = ensure_bytes(payload)
-    padded_size = padded_len(byte_payload)
-    return padded_size, byte_payload.ljust(padded_size, b'\x00')
 
 
 def bytelen(item):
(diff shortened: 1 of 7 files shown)


caproto/_commands.py
```diff
@@ -606,20 +610,26 @@ class SearchRequest(Message):
...
+
+        Maximum length of the channel name (including the terminating) null
+        byte.
     """
     __slots__ = ()
     ID = 6
     HAS_PAYLOAD = True
 
-    def __init__(self, name, cid, version, reply=NO_REPLY):
+    def __init__(self, name, cid, version, reply=NO_REPLY, max_name_len=None):
```
danielballan: I notice that making this configurable on a per-instance basis has the consequence of removing the constraint that an object is fully described by its representation on the wire. This leads to adding the `sentinel` workaround in `safe_repr` and a pretty significant broad loosing of the serialization test.

I'd suggest considering:
1. Does this need to be configurable per instance, or is per-process sufficient?
2. If it does need to be configurable per instance, can the serialization test by tightened to carve out a more specific exemption?

danielballan (commented):
Thanks for the contribution! If I ever knew this about the protocol, I forgot it. :-D

I have one question in line.

caproto/_commands.py
smarsching: I wasn’t sure about the best approach either. At least in a testing context, it might make sense to have this as a per-instance setting and in any case, the socket is needed in order to determine the right number (unless we want to go with the approach by EPICS Base and assume that a UDP packet can always be at least 1024 bytes long).

We could remove the whole length checking from the constructor and instead add a function like `verify_name_len_for_socket` to `_utils.py`. The downside of this is that we would have to convert the `name` from `str` to `bytes` twice, because we need the number of bytes in order to determine whether the length fits.

Maybe, a compromise would be to create a utility function `convert_name` that takes both the `str` and the `socket`, converting the `str` to `bytes` and returning the converted value. We could then pass this value to the constructor as `name`, and inside the constuctor we could check whether `name` is a `str` or `bytes`. In the latter case, we would use it as is (maybe verifying that it is aligned and null-terminated), and in the `str` case, we would convert it using the default limit of 1000 bytes (assuming a maximum UDP packet size of 1024 like EPICS Base does).

What do you think?

danielballan: I like that suggestion! It reminds me of [Parse, Don't Validate](https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/).

smarsching: I rewrote the whole patch as discussed. In order to avoid duplicating code, I moved `padded_len` and `padded_string_payload` from `_commands.py` to `_utils.py`. I think that this makes the changes more localized and compact, as demonstrated by the fact that the modifications to the test code are minimal now.

danielballan (approved):
Thanks! I haven't looked into what's wrong with the CI, but the change set looks good to me.

danielballan:
I see this failed during test setup with:

```
CondaHTTPError: HTTP 429 Too Many Requests for url <[https://conda.anaconda.org/conda-forge/linux-64/repodata.json>](https://conda.anaconda.org/conda-forge/linux-64/repodata.json%3E)
```

I saw that on conda-forge earlier today as well. I think there must be a transient issue with `conda.anaconda.org` at the moment. Let's re-run next week.

smarsching:
@danielballan Do you know how we can trigger the GitHub actions for the tests again, so that we can see whether they pass now?

danielballan:
Power-cycled to re-run CI

danielballan:
There are some true failures in these CI logs now.

smarsching:
Correct. There was a `payload % 8` where `len(payload) % 8` would have been correct. I amended the commit and force-pushed it. Let’s see whether the tests pass now.

smarsching:
Some tests still fail, but I think that this is unrelated to this changeset and rather caused by a problem in `caproto/tests/test_cli_scripts.py`:

```py
@pytest.mark.parametrize('more_kwargs,',
                         [{'repeater': False},
                          {'timeout': 3},
                          {'notify': True},
                          ]
                         )
@pytest.mark.parametrize('func,args,kwargs',
                         [(read, ('float',), {}),
                          (write, ('float', 3.16), {}),
                          (write, ('float', '3.16'), {'data_type': 0}),
                          ])
def test_options(func, args, kwargs, more_kwargs, ioc):
```

It seems that the first `@pytest.mark.parametrize` annotation is not handled correctly, and the strings in the dicts are treated as test names (at least, the output kind of implies that). Interestingly, this only seems to happen in Python 3.10 and newer. Maybe this problem was introduced with a certain version of pytest?

Anyway, I am quite confident that this issue is independent of this patch, so it should probably be addressed by a separate PR.

danielballan:
Thanks, @smarsching!

